PR #8, opened 2021-04-09
tar/import: use new writing APIs

cgwalters:
This is a lot more efficient; before we were creating a thread
per object, etc.

### DIFF
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -5,7 +5,11 @@ use crate::Result;
 use anyhow::{anyhow, Context};
 use camino::Utf8Path;
 use fn_error_context::context;
+use gio::prelude::*;
+use glib::Cast;
+use ostree::ContentWriterExt;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::io::prelude::*;
 
 /// Arbitrary limit on xattrs to avoid RAM exhaustion attacks. The actual filesystem limits are often much smaller.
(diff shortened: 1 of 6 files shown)


cgwalters:
Depends https://github.com/ostreedev/ostree-rs/pull/2

lib/src/tar/import.rs
```diff
@@ -16,6 +20,9 @@ const MAX_XATTR_SIZE: u32 = 1024 * 1024;
 /// from ostree-core.h.  TODO: Bind this in introspection
 const MAX_METADATA_SIZE: u32 = 10 * 1024 * 1024;
 
+/// https://stackoverflow.com/questions/258091/when-should-i-use-mmap-for-file-access
+const SMALL_REGFILE_SIZE: usize = 1; // 127 * 1024;
```
cgwalters, line 24: TODO uncomment

cgwalters:
Hmm actually seeing a failure now if we flip over to using the inline write API:
```
---- test_tar_import_export stdout ----
Error: Importing

Caused by:
    0: object 14/0e43bbd6b53484d27e1dfc4454470d009238b4417080758b7a1676706c89df.file
    1: Processing content object 140e43bbd6b53484d27e1dfc4454470d009238b4417080758b7a1676706c89df
    2: Writing content object: Corrupted file object; checksum expected='140e43bbd6b53484d27e1dfc4454470d009238b4417080758b7a1676706c89df' actual='3c0a99b2e55b312a8a092e632bc9187da441351108bc61a8f9102e6310e9e391'
thread 'test_tar_import_export' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /var/home/walters/.rustup/toolchains/1.48-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:192:5
```

cgwalters:
OK CI failed as expected to due dep on just-released ostree.  (Which is good, I wanted to see it fail to know we're actually testing things :wink: )  Added a hack to update CI to the latest.

cgwalters:
OK cool, CI is passing now.

Issue #6, opened 2021-04-06
container: Support forking off podman


cgwalters:
The current prototype pull code uses https://github.com/deislabs/krustlet/blob/master/crates/oci-distribution/

Advantages:

- It's a thin wrapper for making HTTP requests and is already in Rust
- Due to lack of features, it is very predictable

Disadvantages:

- No support for `~/.docker/config.json` directly

Instead, we could try e.g. forking off podman or skopeo as a subprocess.

Advantages:

- We can honor things like image content source policy
- We'd reuse the code for parsing `~/.docker/config.json` etc.

Disadvantages:

- Forking a subprocess and parsing the output is always eww
- I am not offhand seeing a CLI to stream the output that exists today; we'd need to add it

If we add support for the forking route, I think we should try to sandbox it when running fully privileged.  Let's ensure that it works to do it under e.g. `systemd-run -p DynamicUser=yes` so we know that no persistent state is written.

cgwalters:
I took a stab at this in https://github.com/containers/podman/pull/10075 and https://github.com/ostreedev/ostree-rs-ext/tree/backend-podman but:

- There's some sort of bug that causes the hyper http code to hang and that was a rough introduction to debugging async rust
- We should have skopeo handle "upscaling" the manifests to OCI, but this changes the API needs a bit, and we also need to handle manifest lists
- It may actually work better to have an API request to serialize to `oci-archive` instead of fetching the blobs on our own

So the API might be e.g.:

`/resolve-manifest/quay.io/cgwalters/fcos/latest/linux/amd64` => `sha256:46490f8d5bbe4559b61b9a25724c40a91dc2d20d980b955863b5e47368bfff0a` 

And
`/oci-archive/quay.io/cgwalters/fcos/sha256:46490f8d5bbe4559b61b9a25724c40a91dc2d20d980b955863b5e47368bfff0a` => `oci-archive` tarball.

PR #15, opened 2021-04-22
WIP: Use skopeo to fetch images

cgwalters:
The goal here is to implement: https://github.com/ostreedev/ostree-rs-ext/issues/6

We want to honor things like mirroring set up in `/etc/containers`.

However, after doing all this work I hit on two things:

- containers/image barfs if the destination is a pipe
- And actually even if we ask to copy into an `oci-archive` it still
  spools everything to a temporary directory and then tars it
  back up, entirely obviating the point of streaming.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,3 @@ members = ["cli", "lib"]
 [profile.release]
 codegen-units = 1
 lto = "thin"
-
-[patch.crates-io]
-oci-distribution = { git = 'https://github.com/cgwalters/krustlet', branch = 'streaming-client' }
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 
 [dependencies]
 anyhow = "1.0"
-ostree-ext = { path = "../lib", features = ["container"] }
+ostree-ext = { path = "../lib" }
 clap = "2.33.3"
 structopt = "0.3.21"
 ostree = { version = "0.11.0", features = ["v2021_2"] }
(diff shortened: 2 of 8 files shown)


cgwalters:
For me also this was a *huge* deep dive into async Rust.  I would say I'm intermediate at Rust but I'd only dabbled in async.

In particular the `parse_tar` function is "bridging" the world of async rust to sync because I wanted to keep using the existing sync `tar` crate.   So I spent a *lot* of time trying to figure out how to take an `AsyncRead` and turn that into a plain `Read` usable in a worker thread.  

Then further, I wanted to stream output too, so I ended up using tokio's mpsc channels which are documented to be usable both sync and async.  But even this took me a bit to figure out from searching through the tokio docs.  Web searches for "async sync bridge read" and the like aren't helpful - the terms are too generic. 

Further complicating things I wanted to return an `impl Stream<Item=Result<TarEntry>>` but we also need to poll the future for the helpers.  I tried making my own struct for this but implementing `poll_next()` was a whole other level of learning - particularly around pinning the inner futures, etc.  I ended up just returning two futures. 

In addition to this I also spent a lot of time on inscrutable compiler errors that result when e.g. you forget to include `TryStreamExt` or you forget to wrap the result in a future in a helper closure, etc.  Also some lifetime issues that I am still not sure I quite understand (e.g. my use of the `Rc<String>`), and I ended up tossing in various `.boxed_local()` for lifetime issues on some of the futures.

Now of course there is https://crates.io/crates/tokio-tar but in a very Rust thing, there's an open PR to port to tokio 1.0 that hasn't merged because it's probably just someone's side project or something.

And....after doing all that I very belatedly realized that skopeo doesn't actually stream the results at all :cry: - so all this work on parsing the tar stream is pointless, I could have just forked skopeo to write to a tempdir, and then parse that.  

Ultimately I think we need a more direct containers/image backend to make this truly work well.

PR #16, opened 2021-04-24
container: Use skopeo to fetch images

cgwalters:
Closes: https://github.com/ostreedev/ostree-rs-ext/issues/6

We want to honor things like mirroring set up in `/etc/containers` and ideally
things like signature verification too.  Plus we need to support the Docker
registry API, not pure OCI.

So depending on `oci-distribution` isn't a viable plan from that perspective.

We're not a Go project, so depending on github.com/containers/image directly
is out, plus even if we were vendoring all that is just a bad idea.  So
let's use skopeo as a subprocess.

I originally wrote this to use skopeo via a pipe like:
`skopeo copy docker://quay.io/exampleos/exampleos oci-archive:///proc/self/fd/5`
where fd `5` is a pipe, but the first blocker is that containers/image barfs
if the destination is a pipe.

But even more importantly, the `oci-archive://` backend just
spools everything to a temporary directory and then tars it
back up, entirely obviating the point of streaming.

So here we do the tempdir dance ourself for now.

---


cgwalters:
OK I think this is good to merge.  Anyone want to give it a look over/sanity check, maybe play with the CLI?

cgwalters:
Alright well, since this library is still in an "iron is hot" state and not a dependency of anything yet, in the interest of forward progress here just going to merge.  As it matures we can institute more review coverage (and better CI etc.)

PR #20, opened 2021-04-27
Make tar import API async

cgwalters:
We're currently using the synchronous tar library, so we need to
do bridging.  This pushes the bridging down a layer, changing
our tar import API to be async.

Then things will be even clear if we switch to e.g.
https://crates.io/crates/tokio-tar

But we need https://github.com/vorot93/tokio-tar/pull/3

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -85,15 +85,14 @@ enum Opt {
     Container(ContainerOpts),
 }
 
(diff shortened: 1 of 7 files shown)


lucab (approved):

vorot93:
Sorry for the delay! tokio-tar 0.3.0 targeting tokio 1 has been released.

PR #22, opened 2021-04-28
container: Use docker-archive: to stream image

cgwalters:
Depends: https://github.com/ostreedev/ostree-rs-ext/pull/20 (mostly to avoid conflicts)

container: Use docker-archive: to stream image

It turns out that for whatever reasons (presumably mostly historical)
in skopeo, while `oci-archive:` is backed by just writing everything
to a tempdir and then tarring it up, the `docker-archive:` format
is actually streaming.

And it's *way* nicer and more efficient to stream.  Then the
last piece of this puzzle is that I realized we can use a named pipe
i.e. `mkfifo` to avoid having the containers/image stack crash
because it wants to invoke `realpath()` on the provided target.
By simply giving our pipe a name, it can do so (though it's
pointless) but this way we work with shipped versions of skopeo
and don't need to wait months for patches to propagate.

Also as a bonus, here skopeo ends up decompressing the layer for us,
so we don't need to e.g. handle gzip vs zstd in our code.

### DIFF
--- a/lib/src/container/import.rs
+++ b/lib/src/container/import.rs
@@ -1,9 +1,11 @@
 //! APIs for extracting OSTree commits from container images
 
 use super::*;
-use anyhow::anyhow;
+use anyhow::{anyhow, Context};
+use camino::Utf8Path;
 use fn_error_context::context;
 use futures::prelude::*;
+use std::io::prelude::*;
 use std::process::Stdio;
 use tokio::io::AsyncRead;
 use tracing::{event, instrument, Level};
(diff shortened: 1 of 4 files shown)


cgwalters:
OK rebased :surfer: and lifting draft!

cgwalters:
(OK actually I was wrong it seems, `docker-archive:` still seems to spool the whole uncompressed layer to a file before sending it to us :cry: )

cgwalters:
But to be clear I think it's still worth merging this because we at least re-gain the "incrementally parse streamed tarfile" code which is definitely closer to the desired end state.

lib/src/container/import.rs
```diff
@@ -39,31 +41,120 @@ async fn fetch_manifest(imgref: &ImageReference) -> Result<(oci::Manifest, Strin
...
+        }
+        Err(anyhow!("Failed to find layer {}", blob_symlink_target))
+    })
+    .map_err(anyhow::Error::msg);
+    let stream = tokio_stream::wrappers::ReceiverStream::new(rx_buf);
+    let reader = tokio_util::io::StreamReader::new(stream);
+    // Is there a better way to do this?
+    let worker = async move {
+        let (import, input_copydriver) = tokio::join!(import, input_copydriver);
+        let _: () = import?.context("Import worker")?;
```
kelvinfan001, line 110: Just wondering what the purpose of these unused variables are. Is it just for adding annotations for the compiler?

kelvinfan001 (approved):
Looks sane to me.
(Sorry for lack of any real feedback; this one was a bit harder than I thought 😅, maybe another pair of eyes would help to confirm sanity).

lib/src/container/import.rs
cgwalters: It's confusing because there's `Result<Result<T>>` stuff involved here and I want to be careful we're fully unwrapping these and handling the chain.

It's too easy to do e.g. `let _ = foo?` when it should be `let _ = foo??`.

cgwalters: To elaborate further by doing `let _: ()` and adding the explicit `()` typing we make it a compile error if the result of `?` isn't `()`.

kelvinfan001: makes sense, thanks!

PR #24, opened 2021-04-30
cli+container: Add support for progress

cgwalters:
This demos a lot better than watching the CLI sit there and
do nothing until done!


cgwalters:
https://asciinema.org/a/410974

PR #27, opened 2021-05-11
container/import: Extract layer tar discovery into helper function

cgwalters:
This code was getting too nested and rust-analyzer makes it
very easy to hoist things to helper functions.


kelvinfan001 (approved):
LGTM.

That's pretty cool! I think it's time for me to install rust-analyzer...

PR #39, opened 2021-05-24
Export full CLI in library

cgwalters:
Export full CLI in library

Ultimately I think it will work best if we fully encapsulate
`ostree-ext-cli` in `rpm-ostree` in the short term.  There's
a strong tension in having two binaries.  It *logically* makes
sense, but logistically is harder (duplicating CI, documentation,
crate dependency management, etc.)

Keep the `cli` crate for now, but move all of the functionality
of it into a simple `run_from_iter()` function that can be called
by projects like rpm-ostree.


cgwalters:
Rebased :surfer:

jlebon (approved):
Seems reasonable to me.

travier (approved):

PR #41, opened 2021-05-26
cli: Add `container import --write-ref` option

cgwalters:
This will be needed for https://github.com/coreos/fedora-coreos-tracker/issues/828
so coreos-assembler can find the commit it imported.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -65,6 +65,10 @@ enum ContainerOpts {
 
         /// Image reference, e.g. registry:quay.io/exampleos/exampleos:latest
         imgref: String,
+
+        /// Create an ostree ref pointing to the imported commit
+        #[structopt(long)]
+        write_ref: Option<String>,
     },
 
     /// Print information about an exported ostree-container image.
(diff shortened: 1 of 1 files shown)


lib/src/cli.rs
```diff
@@ -153,19 +157,32 @@ async fn container_import(repo: &str, imgref: &str) -> Result<()> {
...
+            write_ref,
+            Some(import.ostree_commit.as_str()),
+            gio::NONE_CANCELLABLE,
+        )?;
+        println!("Imported: {} => {}", write_ref, import.ostree_commit.as_str());
+    } else {
+        println!("Imported: {}", import.ostree_commit);
     }
+
+    return Ok(())
```
lucab, line 185: Spurious `return` (I feel that the CI here does not run `clippy`).

lucab (approved):

lib/src/cli.rs
cgwalters: Huh I can't get clippy to warn about that.  You're right CI isn't, I'll do a followup PR to add that and fix the existing issues.

cgwalters: Fixed

PR #45, opened 2021-06-03
oci: Generate correct uncompressed hash

cgwalters:
oci: Add crate name/version in history

For debugging purposes.

---

oci: Generate correct uncompressed hash

I broke this in a refactoring, turns out `docker` checks it
but `podman` doesn't (which seems to clearly be a bug).

Reported on chat by someone trying out running
via `docker`.

---

### DIFF
--- a/lib/src/container/oci.rs
+++ b/lib/src/container/oci.rs
@@ -312,6 +312,6 @@ impl<'a> std::io::Write for LayerWriter<'a> {
     }
 
     fn flush(&mut self) -> std::io::Result<()> {
-        Ok(())
+        self.bw.flush()
     }
 }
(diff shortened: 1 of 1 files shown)


lucab:
If you are not in a hurry, I'd really consider adding some basic unit tests in this area because it has good chances to see further mistakes/regressions. But the patch itself looks good.

lucab (approved):

gicmo:
> Reported on chat by someone trying out running via docker.

Heh.

FWIW, the code looks good to me.

cgwalters:
> If you are not in a hurry, I'd really consider adding some basic unit tests in this area because it has good chances to see further mistakes/regressions.

The code here is indirectly tested via the higher level code FWIW, but I did add a basic unit test that covers the uncompressed hash.

PR #50, opened 2021-07-14
container/import: Only close skopeo fifo once the process has exited

cgwalters:
We're hitting this race condition when teaching rpm-ostree to
consume container images.  The current skopeo code creates
two async tasks, one to read from the FIFO (pipe), and one
to wait on the child process.

Currently when we find the layer tarball we're expecting,
we close the FIFO.  This causes any further messages from
skopeo to fail with `EPIPE` which it's not expecting, which
will then cause skopeo to exit with a failure.

Rework the logic here so that we hold open the FIFO until
the process has exited.  The "tar parsing" import code
takes ownership of the stream, but then passes it back
to the caller.


kelvinfan001 (approved):
Looks sane to me. Tested locally and can confirm this fixes the broken pipe issue.

PR #52, opened 2021-07-24
variant_utils: Add helpers for strv/`as`

cgwalters:
We use this type extensively, so let's add some convenience API.

This should go away with glib 0.14, but that's a nontrivial port.


cgwalters:
Let's do https://github.com/ostreedev/ostree-rs-ext/pull/53 instead

PR #53, opened 2021-07-26
Port to glib 0.14

cgwalters:
Port to glib 0.14

This depends on the WIP port for ostree-rs.

---

Update to use new GLib variant bindings

This drops some unsafe code and increases ergonomics.  I'd like
to kill our `variant_utils` entirely but that will still need
more work.

---


cgwalters:
Depends https://github.com/ostreedev/ostree-rs/pull/7

cgwalters:
OK published new https://crates.io/crates/ostree 0.12 and switched to use that here!

lib/src/variant_utils.rs
```diff
@@ -32,36 +32,21 @@ pub fn variant_normal_from_bytes(ty: &str, bytes: glib::Bytes) -> glib::Variant
...
-        let r = glib_sys::g_variant_new_fixed_array(
-            b"y\0".as_ptr() as *const _,
-            buf.as_ptr() as *const _,
-            buf.len(),
-            size_of::<u8>(),
-        );
-        glib_sys::g_variant_ref_sink(r);
-        from_glib_full(r)
-    }
+    buf.to_variant()
```
cgwalters, line 35: Being able to use stuff like this was a primary motivation for the bump - look ma no `unsafe`!

cgwalters: And done!  Almost all of `variant_utils.rs` is gone, and the only `Variant` bits are just `pub(crate)`.  There's still `VariantDictExt` which I think is kind of still useful...needs some investigation.

lib/src/ima.rs
```diff
@@ -47,10 +48,19 @@ fn xattrs_to_map(v: &glib::Variant) -> BTreeMap<Vec<u8>, Vec<u8>> {
     map
 }
 
-/// Reserialize a map to GVariant of type `a(ayay)`
-fn xattrmap_serialize(map: &BTreeMap<Vec<u8>, Vec<u8>>) -> glib::Variant {
-    let map: Vec<_> = map.iter().collect();
-    variant_utils::new_variant_a_ayay(&map)
+/// Create a new GVariant of type a(ayay).  This is used by OSTree's extended attributes.
+fn new_variant_a_ayay<'a, T: 'a + AsRef<[u8]>>(
```
cgwalters, line 52: Not gonna lie after I wrote this I felt like I'd just inscribed a new page in a spellbook.

lib/src/tar/import.rs
```diff
@@ -434,7 +433,7 @@ impl<'a> Importer<'a> {
         let mut contents = vec![0u8; n as usize];
         entry.read_exact(contents.as_mut_slice())?;
         let contents: glib::Bytes = contents.as_slice().into();
-        let contents = variant_new_from_bytes(OSTREE_XATTRS_FORMAT, contents, false);
+        let contents = Variant::from_bytes::<&[(&[u8], &[u8])]>(&contents);
```
cgwalters, line 436: Note how we were previously passing a string `a(ayay)` and now encode the Variant type via Rust generics.

lib/src/ima.rs
```diff
@@ -47,10 +48,19 @@ fn xattrs_to_map(v: &glib::Variant) -> BTreeMap<Vec<u8>, Vec<u8>> {
...
+) -> glib::Variant {
+    let children: Vec<_> = items
+        .into_iter()
+        .map(|(a, b)| {
+            let a = a.as_ref();
+            let b = b.as_ref();
+            Variant::from_tuple(&[a.to_variant(), b.to_variant()])
+        })
+        .collect();
+    Variant::from_array::<(&[u8], &[u8])>(&children)
```
lucab, line 63: Self-note: this internally gets ownership of the content, see note at https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-new-array.

lucab (approved):

PR #56, opened 2021-07-28
cargo/lib: update all dependencies

lucab:
This sorts the dependencies in 'lib' manifest, updates them all to
the latest versions, and massages code for all changed APIs.


lucab:
@cgwalters I don't have access to the settings tab here, can you please check what is the Dependabot status on this repo and why the dependencies are not getting updated?

cgwalters (approved):
Nice!

I enabled Dependabot here.

PR #58, opened 2021-07-30
oci: Pass through other architectures

cgwalters:
I think the OCI arch matches the kernel architecture on other
platforms.


lucab:
The current list is at https://golang.org/doc/install/source#environment. This PR should enough to at least take care of `ppc64le` and `s390x`.

lucab (approved):

PR #60, opened 2021-08-03
lib: Drop gvariant_utils module

cgwalters:
We can now reuse the main GLib 0.14 code since we have the
binding for `g_variant_get_normal_form()`.


cgwalters:
This will depend on both a new ostree and glib version.

cgwalters:
Soft requires https://github.com/ostreedev/ostree-rs/pull/26

cgwalters:
Lifting draft since https://crates.io/crates/glib/0.14.4 is out!

jlebon (approved):
Nice!

PR #63, opened 2021-08-05
Try to fix the docs.rs build

cgwalters:
I am totally cargo culting the bits from ostree-rs without understanding
them.  I guess this somehow avoids having the dependencies installed
in the random docs.rs build container?

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -43,3 +43,9 @@ clap = "2.33.3"
 indoc = "1.0.3"
 sh-inline = "0.1.0"
 structopt = "0.3.21"
+
+[package.metadata.docs.rs]
+features = ["dox"]
+
+[features]
+dox = ["ostree/dox"]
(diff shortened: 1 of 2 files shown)


lucab:
Yes! Specifically, not requiring the latest version to document everything. See https://gtk-rs.org/blog/2017/11/26/new-release.html.

lib/Cargo.toml
```diff
@@ -43,3 +43,6 @@ clap = "2.33.3"
 indoc = "1.0.3"
 sh-inline = "0.1.0"
 structopt = "0.3.21"
+
+[package.metadata.docs.rs]
+features = ["dox"]
```
lucab, line 48: I think you will have to chain this to the other dependencies, like https://github.com/gtk-rs/gtk-rs-core/blob/0.14.2/gio/Cargo.toml#L31.

cgwalters: Thanks, fixed!

cgwalters:
OK I verified this works locally by doing:
```
$ yum -y remove ostree-devel
$ cargo +nightly doc
...
<failure due to missing headers>
$ cargo +nightly doc --features=dox
...
<success>
$
```

lucab (approved):

PR #70, opened 2021-08-27
 Depend on oci-spec

cgwalters:
Since https://github.com/containers/oci-spec-rs now
exists, let's use it!


cgwalters:
Ah, oci-spec-rs is depending on https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1540-2021-07-29 from Rust 1.54.  Hmm.

PR #72, opened 2021-09-09
Sign verify tar prep

cgwalters:
tar/import: Minor state machine cleanup

Validate upfront that we only get xattrs with content objects;
this way the rest of the match doesn't need to ignore the boolean.

---

tar: Validate we get an error with an empty tarball

And then we can later extend this to more error cases.

---

### DIFF
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -351,27 +351,26 @@ impl<'a> Importer<'a> {
         };
         let objtype = objtype_from_string(objtype)
             .ok_or_else(|| anyhow!("Invalid object type {}", objtype))?;
-        match (objtype, is_xattrs, &self.state) {
-            (ostree::ObjectType::Commit, _, ImportState::Initial) => {
+        if is_xattrs && objtype != ostree::ObjectType::File {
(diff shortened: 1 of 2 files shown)


cgwalters:
I have like 5 local branches of WIP-then-abandoned bits for tar+signing :cry:

lib/src/tar/import.rs
```diff
@@ -351,28 +351,27 @@ impl<'a> Importer<'a> {
         };
         let objtype = objtype_from_string(objtype)
             .ok_or_else(|| anyhow!("Invalid object type {}", objtype))?;
-        match (objtype, is_xattrs, &self.state) {
-            (ostree::ObjectType::Commit, _, ImportState::Initial) => {
+        if is_xattrs && objtype != ostree::ObjectType::File {
+            return Err(anyhow!("Found xattrs for non-file object type {}", objtype));
```
lucab, line 355: Here and below, I think you were looking for `bail!()`.

lucab (approved):

lib/src/tar/import.rs
cgwalters: I kind of prefer the explicitness of seeing `return` actually myself.

PR #74, opened 2021-09-09
lib: Only depend on futures-util

cgwalters:
I saw this in https://www.reddit.com/r/rust/comments/pkr9aa/is_the_crate_dependency_becoming_a_problem/

That said it turns out `glib` is depending on `futures-executor` right
now too.


lucab (commented):
I have the feeling we are only using a few traits from `futures-utils`. If that's the case, it would be good to name them explicitly and remove the star-import.

cgwalters:
Mmmm...ok done.

lucab (approved):

PR #75, opened 2021-09-10
Tar cleanup 2

cgwalters:
lib/tar/import: Move more logic into Importer struct

Let's keep the code inside an anonymous closure smaller.

---

lib/tar/import: Remove lifetime from Importer

It's really easier to just bump the gobject refcount; not
worth carrying a lifetime for this.

---

lib/tar/import: Explicitly parse commit object first

Prep for signing.  Rather than maintaining a state machine,
just explicitly read the first object and validate it as
a commit.

This way we can later also read the commitmeta in the same
way and validate both before writing them to the repo.

---

### DIFF
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -44,9 +44,9 @@ struct ImportStats {
 }
 
 /// Importer machine.
-struct Importer<'a> {
+struct Importer {
     state: ImportState,
-    repo: &'a ostree::Repo,
+    repo: ostree::Repo,
     xattrs: HashMap<String, glib::Variant>,
     next_xattrs: Option<(String, String)>,
 
@@ -56,7 +56,7 @@ struct Importer<'a> {
     stats: ImportStats,
 }
 
-impl<'a> Drop for Importer<'a> {
+impl Drop for Importer {
     fn drop(&mut self) {
         let _ = self.repo.abort_transaction(gio::NONE_CANCELLABLE);
     }
@@ -115,11 +115,11 @@ fn entry_to_variant<R: std::io::Read, T: StaticVariantType>(
     Ok(v.normal_form())
 }
 
-impl<'a> Importer<'a> {
-    fn new(repo: &'a ostree::Repo) -> Self {
+impl Importer {
+    fn new(repo: &ostree::Repo) -> Self {
         Self {
             state: ImportState::Initial,
-            repo,
+            repo: repo.clone(),
             buf: vec![0u8; 16384],
             xattrs: Default::default(),
             next_xattrs: None,
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -3,6 +3,7 @@
 use crate::Result;
 use anyhow::{anyhow, Context};
 use camino::Utf8Path;
+use camino::Utf8PathBuf;
 use fn_error_context::context;
 use futures_util::TryFutureExt;
 use gio::glib;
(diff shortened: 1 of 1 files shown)


cgwalters:
What you don't see here is the like 4-5 prior broken/ugly branches that preceeded this.  As well as the solid 1.5 hours I spent chasing a crazy mysterious bug where the tar entry `.path()` was returning something totally unexpected.  I finally narrowed it down to this:

```
diff --git a/lib/src/tar/import.rs b/lib/src/tar/import.rs
index 01ba3cd..a417e58 100644
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -158,7 +158,7 @@ impl Importer {
     fn filter_entry<R: std::io::Read>(
         e: tar::Entry<R>,
     ) -> Result<Option<(tar::Entry<R>, Utf8PathBuf)>> {
-        let orig_path = e.header().path()?;
+        let orig_path = e.path()?;
         let path = Utf8Path::from_path(&*orig_path)
             .ok_or_else(|| anyhow!("Invalid non-utf8 path {:?}", orig_path))?;
         // Ignore the regular non-object file hardlinks we inject
```

And indeed, the docs say:
> It is recommended to use this method instead of inspecting the header directly to ensure that various archive formats are handled correctly.

But what I was seeing is that the `e.header().path()` somehow returned the path to the *previous* parsed entry and this had me thinking memory corruption was involved, compiler bugs...but I guess calling `e.header().path()` bypasses important parts of the tar parsing state machine or something?  Anyways, that was painful.

lib/src/tar/import.rs
```diff
@@ -115,11 +115,11 @@ fn entry_to_variant<R: std::io::Read, T: StaticVariantType>(
     Ok(v.normal_form())
 }
 
-impl<'a> Importer<'a> {
-    fn new(repo: &'a ostree::Repo) -> Self {
+impl Importer {
+    fn new(repo: &ostree::Repo) -> Self {
```
lucab, line 145: I do wonder whether it would be better to take full ownership of the `ostree::Repo` argument from the caller here, and give it back through a new `finish()` method when importing is over. It would also be more in line with usual Rust rules, as the immutable borrow here is a bit of lie (importing steps do mutate the repo). How are your feelings on that?

lib/src/tar/import.rs
```diff
@@ -115,10 +109,41 @@ fn entry_to_variant<R: std::io::Read, T: StaticVariantType>(
     Ok(v.normal_form())
 }
 
+fn parse_object_entry_path(path: &Utf8Path) -> Result<(&str, &Utf8Path, &str)> {
+    let parentname = path
+        .parent()
+        .map(|p| p.file_name())
+        .flatten()
+        .ok_or_else(|| anyhow!("Invalid path (no parent) {}", path))?;
+    if parentname.len() != 2 {
```
lucab, line 118: It took a me a bit to grasp what is this 2 here.
My understanding is that this is checking for the sharding-prefix of the object checksum as the parent directory (sorry, I don't know if this entity has a proper name), am I reading it right?
A brief comment here would be helpful for future readers.

lucab (approved):
LGTM. Two minor readability/style doubts, but they can addressed later on.

lib/src/tar/import.rs
cgwalters: Right; gtk-rs made a core decision that GObjects always have interior mutability.  Because it's really the 90% case.  And it does indeed mean that one can get memory corruption if invoking methods on a gobject from multiple threads if those objects/methods aren't explicitly threadsafe internally.

So I think trying to do such a dance wouldn't really help much.

lib/src/tar/import.rs
cgwalters: OK, added a comment.

PR #77, opened 2021-09-13
lib/container: Introduce OstreeImageReference

cgwalters:
lib/container: Introduce OstreeImageReference with SignatureSource

I'm trying to put my foot down and say that *by default* booting
and updating an operating system should require a cryptographic
signature from the vendor as implemented by e.g. ostree (GPG or
"signapi" which can do simple ed25519+libsodium signatures).

Other signature mechanisms are possible and the containers/image
ecosystem supports signatures, but they are not widely deployed.

In order to implement ostree-based GPG/signapi verification, we
need the configuration for an ostree remote.  Hence, we support
`ostree-remote-image:$remotename:$containerref`.  For example,
`ostree-remote-image:fedora:registry:quay.io/coreos/fedora-coreos:stable`.

Having a canonical stringified representation of this will
allow using it on command lines and in user interfaces.  For example:

```
$ ostree remote add fedora --set=gpgkeypath=/etc/pki/rpm-gpg --custom-backend=ostree-ext
$ ostree-ext-cli rebase ostree-remote-image:fedora:quay.io/coreos/fedora-coreos:stable
```

(`ostree-ext-cli rebase` doesn't exist, but maybe it will later)

In addition, a signature policy of `ostree-image-signed` expresses
that we won't do signature verification via ostree, but we *will*
try to ensure that the containers/image (skopeo) path uses some
signature mechanism.

Finally we offer `ostree-unverified-image` to just boot a container
without caring about signatures.

---

container: Add support for `ostree-remote-registry`

This is a convenient shorthand for the very common case of fetching
from a registry (i.e. `docker://`).

---

### DIFF
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -40,12 +40,35 @@ pub enum Transport {
 /// Combination of a remote image reference and transport.
 ///
 /// For example,
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ImageReference {
     /// The storage and transport for the image
     pub transport: Transport,
     /// The image name (e.g. `quay.io/somerepo/someimage:latest`)
     pub name: String,
+}
+
+/// Policy for signature verification.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SignatureSource {
+    /// Fetches will use the named ostree remote for signature verification of the ostree commit.
+    OstreeRemote(String),
+    /// Fetches will defer to the `containers-policy.json`, but we make a best effort to reject `default: insecureAcceptAnything` policy.
+    ContainerPolicy,
+    /// NOT RECOMMENDED.  Fetches will defer to the `containers-policy.json` default which is usually `insecureAcceptAnything`.
+    ContainerPolicyAllowInsecure,
+}
+
+/// Combination of an ostree remote (for signature verification) and an image reference.
+///
+/// For example, myremote:docker://quay.io/somerepo/someimage.latest
+#[derive(Debug, Clone)]
+pub struct OstreeImageReference {
+    /// The ostree remote name.
+    /// This will be used for signature verification.
+    pub sigverify: SignatureSource,
+    /// The container image reference.
+    pub imgref: ImageReference,
 }
 
 impl ImageReference {
@@ -69,6 +92,23 @@ impl ImageReference {
         Self {
             transport: self.transport,
             name: format!("{}@{}", name, digest),
+        }
+    }
+}
+
+impl OstreeImageReference {
+    /// Create a new `OstreeImageReference` that refers to a specific digest.
+    ///
+    /// ```rust
+    /// use std::convert::TryInto;
+    /// let r: ostree_ext::container::OstreeImageReference = "ostree-remote-image:myremote:docker://quay.io/exampleos/exampleos:latest".try_into().unwrap();
+    /// let n = r.with_digest("sha256:41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3");
+    /// assert_eq!(n.imgref.name, "quay.io/exampleos/exampleos@sha256:41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3");
+    /// ```
+    pub fn with_digest(&self, digest: &str) -> Self {
+        Self {
+            sigverify: self.sigverify.clone(),
+            imgref: self.imgref.with_digest(digest),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


cgwalters:
Trying to teach `rpm-ostree rebase` about the new signing stuff, I hit ergnomic issues.   This is just one idea.  

Another one is to make this a URL, like `ostreeregistry:myremote:quay.io/exampleos:latest`.  So `ostreeregistry` is a scheme (like `https`) and its syntax is `REMOTENAME:IMGREF`.   Actually I'm leaning towards that.

cgwalters:
Maybe `ostreeimage`?  Or something else?

cgwalters:
OK I reworked this per https://github.com/ostreedev/ostree-rs-ext/pull/77#issuecomment-918430662 but still need to implement `SignatureSource::ContainerPolicy` (and add an API that uses `OstreeImageReference` but that's easy).

cgwalters:
OK, lifting draft on this one.  I only lightly tested the `SignaturePolicy::ContainerPolicy` and `SignaturePolicy::ContainerPolicyAllowInsecure` paths, but I don't think it's a really big deal right now, just a best effort from our part at this point.  The important one is `SignaturePolicy::OstreeRemote`.

lucab:
Just double-checking, `:` is a disallowed character in an OSTree remote name, right? (I haven't encountered it before, but I think [this](https://github.com/ostreedev/ostree/blob/v2021.4/src/libostree/ostree-core.c#L151-L153) should be the relevant grammar).
If I'm understanding this scheme proposal correctly, it does rely on `:` being a reserved/separator token which can't appear in the inner remotename values.

cgwalters:
Yep, that's exactly right.  Kelvin and I debated this extensively.  It's why today doing:

`rpm-ostree rebase --experimental registry:quay.io/coreos/fedora-coreos`
fails today because it's ambiguous whether `registry` is an ostree remote name or a containers-image schema.

(`registry` and `docker://` are synonymous to containers-image)

This PR is basically introducing a much more verbose and not-ambiguous syntax to explicitly denote we're fetching an ostree-container. 

Which, cuts against the "container native" bit by introducing something new and unique to us.  But OTOH, booting a container *is* special, and the extra typing/cognitive burden shouldn't be a problem in practice.

cgwalters:
See also e.g. https://github.com/coreos/coreos-assembler/pull/2402

cgwalters:
Other possibilities:

`ostree-remote-registry:fedora:quay.io/coreos/fedora:stable`

(i.e. we add a shorthand `remote-registry` for `remote-image:...:registry:...`)

An entirely different one is something like: `quay.io/coreos/fedora:stable?signature=ostree:fedora` - here we extend the container image FQIN with a HTTP-like query parameter syntax.  The danger here is a potential future incompatibility with an extension to container image FQIN.

cgwalters:
> (i.e. we add a shorthand remote-registry for remote-image:...:registry:...)

Did this.  But there's still the open question about other formats.

lib/src/container/skopeo.rs
```diff
@@ -2,9 +2,42 @@
 
 use super::Result;
 use anyhow::Context;
+use serde::Deserialize;
 use std::process::Stdio;
 use tokio::process::Command;
 
+const POLICY_PATH: &str = "/etc/containers/policy.json";
```
lucab, line 9: I think the types here below are mirroring https://github.com/containers/image/blob/main/signature/policy_types.go, let's add a direct link to it.

lib/src/container/mod.rs
```diff
@@ -112,6 +154,69 @@ impl TryFrom<&str> for ImageReference {
...
+            "ostree-remote-registry" => {
+                let mut subparts = second.splitn(2, ':');
+                // Safety: Split always returns at least one value.
+                let remote = subparts.next().unwrap();
+                let rest = subparts
+                    .next()
+                    .ok_or_else(|| anyhow!("Missing second ':' in {}", value))?;
+                (
+                    SignatureSource::OstreeRemote(remote.to_string()),
+                    Cow::Owned(format!("registry:{}", rest)),
```
lucab, line 197: I think we may end up with a `registry:docker://foo` at this point, which I think could become problematic down the road. Also, it would be good to have a testcase covering that.

lib/src/container/mod.rs
```diff
@@ -112,6 +154,69 @@ impl TryFrom<&str> for ImageReference {
...
+                        .next()
+                        .ok_or_else(|| anyhow!("Missing second ':' in {}", value))?,
+                );
+                (SignatureSource::OstreeRemote(remote.to_string()), second)
+            }
+            o => {
+                return Err(anyhow!("Invalid signature source: {}", o));
+            }
+        };
+        let imgref = rest.deref().try_into()?;
```
lucab, line 215: I think we should add some checks to guarantee that this starts with exactly one of `registry:` or `docker://`. Or maybe strip the scheme completely in-memory and add a formatter to print it in a uniform way.

lucab:
I do like the shorter `ostree-remote-registry`, I didn't see ambiguities. Are you maybe thinking of fully dropping `ostree-remote-image` in favor of that, or there will be consumers needing that?

cgwalters:
> Are you maybe thinking of fully dropping ostree-remote-image in favor of that, or there will be consumers needing that?

I think we need to support e.g. `ostree-remote-image:fedora:oci-archive:/path/to/some.ociarchive` and `ostree-remote-image:fedora:containers-storage:quay.io/coreos/fedora:stable` - i.e. fetching from not-docker-registry sources in the same way `skopeo` and the https://github.com/containers/image supports.

cgwalters:
We should probably also have a shorthand `ostree-registry-signed:quay.io/coreos/stable` that expands to `ostree-image-signed:registry:quay.io/coreos/stable` or so.

cgwalters:
Also @travier pointed out right now we'll be discarding any container image signatures during pull.  It'd be nicer to have skopeo also spit these out and save them so we can show that (and re-execute the signature verification again).  Alternatively, we can just serialize the data for signature verification and show that (i.e. don't re-execute verification).

lib/src/container/mod.rs
cgwalters: This and your above point both require us to parse the image reference.  Which is doable but a not-small scope increase, because there are a lot of "transports" supported by skopeo and containers/image, e.g. `containers-storage:` and `oci-archive:` are also both used.

In the end, a malformed image reference is going to result in an error (from skopeo) - we'd just be moving that error up earlier.  (Which is valuable, but IMO not critical)

ashcrow:
:+1: I like this idea and how it's shaping up!

cgwalters:
OK, let's roll with this for 0.3 - in rpm-ostree this functionality is still experimental, and I think anyone else using this crate is probably aware that it's still evolving and may change, so a semver bump to 0.4 in the future is likely, and maybe we end up with different strings.

But I'd like to try to stick to this unless there's a decently strong reason to break it.

PR #81, opened 2021-09-15
tar+container rustdoc

cgwalters:
lib/{tar,container}: Add a lot more module level docs

This overlaps with the `README.md` but I plan to tweak things
in a bit to have that link to `docs.rs`.

---

lib/cli: Fix link to clap

Because we're not importing `clap` directly but only indirectly via
`structopt`, we got a warning from `cargo doc`.

---


lucab (approved):
This was a great read and would be immensely useful for references, thanks for writing down all the inner details as prose!

PR #84, opened 2021-09-16
Bump to 0.4, various container import API changes

cgwalters:
lib: Bump to 0.4.0-alpha.0

I plan to make some API changes in the `container` module at least.

---

lib/container: Change API to use complete digested image references

Rather than having our manifest fetch API just return a `String`
that's a sha256 checksum, instead return a full `OstreeImageReference`
copy of the input image reference, just with a digest.

This helps ensure we're consistently using image references.

---

lib/container: Only have one manifest fetch path

It was confusing to have a codepath that returns just the
digest, and another that fetches the parsed data and the digest.

Instead, fetch *unparsed* data (so that we don't need to expose
`oci::Manifest` as public API) as well as the digested image.

Prep for avoiding an extra round trip in the import path
to fetch the manifest when the caller already has it, in case
they wanted to do change detection.

---

lib/container: Add an API that uses an already fetched manifest

The flow here is then nicer, because at a high level apps like
rpm-ostree will want to do:

- Fetch manifest, digest pair
- Do we already have this? If so, we're done
- Otherwise, perform the fetch using that already downloaded manifest

Previously we were fetching the manifest twice.

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.3.1-alpha.0"
+version = "0.4.0-alpha.0"
 
 [dependencies]
 anyhow = "1.0"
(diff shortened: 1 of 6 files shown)


cgwalters:
OK yeah it turns out my ambition to try to suffix consistently with `@sha256:` to communicate digests doesn't work; it's not supported by `oci:`, `oci-archive:` for example.  This is all pretty clear in `man containers-transports`.  Blah.

PR #85, opened 2021-09-17
container/import: Check skopeo errors first

cgwalters:
Right now a failure during skopeo pull when ostree-remote verification
is enabled actually manifests as "no commit found".

Prioritize emitting a skopeo error, because a failure there probably
has the real error.


lucab (approved):

cgwalters:
BTW, same procedural question/hope here from https://github.com/coreos/coreos-assembler/pull/2440#issuecomment-920887906

PR #88, opened 2021-09-17
Digest cleanups 2

cgwalters:
Depends https://github.com/ostreedev/ostree-rs-ext/pull/87

---

lib/container: Change manifest fetch API to return raw bytes and digest

This allows us to use it more consistently.  We also drop the
unnecessary wrapper structure.

---

lib/container: Add API to fetch from already downloaded manifest

The flow here is then nicer, because at a high level apps like
rpm-ostree will want to do:

- Fetch manifest, digest pair
- Do we already have this? If so, we're done
- Otherwise, perform the fetch using that already downloaded manifest

Previously we were fetching the manifest twice.

---

lib/container: Make use of `skopeo copy --digestfile` if present

This came from https://github.com/containers/skopeo/pull/1266
and has landed in Fedora 34 at least.

---


cgwalters:
Rebased :surfer: and lifting draft!

lucab (approved):

PR #90, opened 2021-09-20
lib/container: Explicitly drain stream

cgwalters:
Rather than passing the FIFO stream ownership into and back out
of the parser, just pass an `&mut` reference, and then
explicitly drain it in the worker thread.

This motivates then cleaning up our "worker" futures.  Add some
comments too.

Prep for future work.

---

### DIFF
--- a/lib/src/container/import.rs
+++ b/lib/src/container/import.rs
@@ -118,36 +118,30 @@ async fn fetch_manifest(imgref: &OstreeImageReference) -> Result<(oci::Manifest,
 
 /// Read the contents of the first <checksum>.tar we find.
 /// The first return value is an `AsyncRead` of that tar file.
-/// The second return value is a background worker task that will
-/// return back to the caller the provided input stream (converted
-/// to a synchronous reader).  This ensures the caller can take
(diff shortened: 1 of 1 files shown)


lib/src/container/import.rs
```diff
@@ -125,18 +125,19 @@ async fn fetch_manifest(imgref: &OstreeImageReference) -> Result<(oci::Manifest,
 pub async fn find_layer_tar(
     src: impl AsyncRead + Send + Unpin + 'static,
     blobid: &str,
-) -> Result<(
-    impl AsyncRead,
-    impl Future<Output = Result<impl Read + Send + Unpin + 'static>>,
-)> {
+) -> Result<(impl AsyncRead, impl Future<Output = Result<()>>)> {
```
lucab, line 128: The docstring needs to be update to match the new signature/logic.

lucab (approved):

lib/src/container/import.rs
cgwalters: OK, I cleaned up the related logic here too.

lucab (approved):

PR #93, opened 2021-09-20
tests: Verify that exported tar is bit-for-bit identical

cgwalters:
It's handy to have some level of predictability, so verify
that the tar stream we output for a given commit is
bit-for-bit identical when invoked twice - with a one
second delay to flush out at least things like modification
times.

We will never be able to assert on a particular checksum, because
e.g. things like the gzip algorithm may compress differently
when the code changes.  Related discussion in
https://github.com/cgwalters/git-evtag#tarball-reproducibility

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -143,6 +143,28 @@ async fn test_tar_import_empty() -> Result<()> {
 }
 
 #[tokio::test]
+async fn test_tar_export_reproducible() -> Result<()> {
+    let fixture = Fixture::new()?;
+    let (_, rev) = fixture
+        .srcrepo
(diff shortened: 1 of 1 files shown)


lucab (approved):
I'm not seeing the artificial delay you mention, but at the same time I don't think that's really need when exporting. Or did you maybe want to pause after fixtures creation?

cgwalters:
> I'm not seeing the artificial delay you mention, but at the same time I don't think that's really need when exporting. 

Yeah, in this case it really shouldn't be necessary, but that's part of what we're testing!  Re-added (had dropped it in a refactor)

cgwalters:
Also, the container equivalent has things like gzip mtimes which will want this delay.

Issue #18, opened 2021-04-26
container: Work on new skopeo backend


cgwalters:
The way we fork skopeo as a subprocess has a lot of flaws.

See: https://github.com/containers/image/issues/1209

cgwalters:
I think as a PoC, we can make a new sub-project like `ostreedev/ostree-container-fetcher` or something that vendors containers/image and does what we need.

The ideal then would be to fold that code back into `skopeo`; something like e.g. `skopeo registry-proxy` that can speak HTTP over a unix domain pipe.  We still want it to do things like signature verification and layer digest verification, but otherwise fetching and parsing the blobs is via us doing pulls.

cgwalters:
https://github.com/cgwalters/container-image-proxy now exists

cgwalters:
Some extremely hacky WIP on this in https://github.com/cgwalters/ostree-rs-ext/commits/import-via-proxy

cgwalters:
OK I thought this didn't work, but it turns out it's racy.  The race appears to be somewhere in the Rust async code around buffering, but I haven't figured out where yet.

PR #94, opened 2021-09-20
lib/container: Convert to oci-archive, not docker-archive:

cgwalters:
The use of `docker-archive:` originally came from
https://github.com/ostreedev/ostree-rs-ext/issues/18
but it turns out I was wrong, skopeo ends up staging everything
anyways.

In practice, what we really want to do here is land custom code
in skopeo to do what we need, instead of treating this as a `copy`
operation, because it isn't.

Relatedly, I also noticed in our `docker-archive:` parsing that we weren't
verifying the relationship of the layer symlink to the sha256 digest;
we were only verifying the *presence* of the symlink.
In practice, skopeo will already be doing the layer verification for
us, but this mechanism is clearer.

Prep for derivation.


cgwalters:
Eh, let's just sprint towards https://github.com/cgwalters/container-image-proxy

PR #96, opened 2021-09-21
lib/tar: Add new `write` module

cgwalters:
APIs to write a tarball stream into an OSTree commit.

This functionality already exists in libostree mostly,
this API adds a higher level, more ergonomic Rust frontend
to it.

In the future, this may also evolve into parsing the tar
stream in Rust, not in C.

Prep for container derivation.

### DIFF
--- a/lib/src/cmdext.rs
+++ b/lib/src/cmdext.rs
@@ -1,0 +1,21 @@
+use std::os::unix::prelude::{CommandExt, RawFd};
+
+pub(crate) trait CommandRedirectionExt {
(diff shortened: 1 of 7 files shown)


cgwalters:
This requires https://github.com/ostreedev/ostree-rs/pull/32 - will look at a new release soon.

cgwalters:
OK released https://crates.io/crates/ostree/0.13.1 so this should be good to go.

lucab (approved):

PR #97, opened 2021-09-21
test: Add an `#[ignore]`d test to verify registry integration

cgwalters:
Right now we're only testing `oci-archive:` really, let's actually
test pushing/pulling via Docker registry API.  Since this requires
external setup, mark it as `#[ignore]`.


lib/tests/it/main.rs
```diff
@@ -361,6 +371,47 @@ async fn test_container_import_derive() -> Result<()> {
     Ok(())
 }
 
+#[ignore]
```
lucab, line 374: For the future, one approach could be to put those tests behind feature flags, see https://github.com/camallo/dkregistry-rs/blob/854d0da53bef5dd85b5e901123e85d43af97c74e/tests/net/quay_io/mod.rs#L16.

lucab (approved):

lib/tests/it/main.rs
cgwalters: OK, hmm yeah could make sense.

PR #100, opened 2021-09-24
Layered import followups

cgwalters:
imageproxy: Add context to failure to spawn

Since missing the binary is probably going to be common.

---

imageproxy: Add some docstrings

On general principle.

---


sohankunkerkar (approved):
LGTM

PR #103, opened 2021-09-27
lib: Lower bitflags requirement to just "1"

cgwalters:
This is dealt with in a newer nix but we need to match
what is locked in rpm-ostree right now.  See
https://github.com/nix-rust/nix/commit/5495bbce52d3541f90d13e692a1cef34e186e100

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -12,7 +12,7 @@ version = "0.4.0-alpha.0"
 anyhow = "1.0"
 async-compression = { version = "0.3", features = ["gzip", "tokio"] }
 bytes = "1.0.1"
-bitflags = "1.3.2"
+bitflags = "1"
 camino = "1.0.4"
 cjson = "0.1.1"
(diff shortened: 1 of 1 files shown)


lucab:
I'm not sure I'm understanding what this PR is fixing. Do we have to support somewhere a toolchain <1.46?

cgwalters:
Or maybe our dependency on bitflags is massive overkill for a single bit :wink:

lucab:
Detailing it through, I was surprised we can't build with both 1.2 and 1.3 in the lockfile.
But then I guess this is because cargo is unifying the 1.x dependency, which is different than what happens for crates at 0.x (which are not compatible and thus do not get unified). 
If we don't have any particular requirement here, I think we can just try specifying `1` so that the final consumer lockfile can pick whatever it wants.

lucab:
Anyway, I'm also fine with just fixing the build now and kicking the can to later improve compatibility, LGTM!

lucab (approved):

cgwalters:
OK yep using just `"1"` works.

PR #104, opened 2021-09-27
lib/tar/write: Translate /etc → /usr/etc by default

cgwalters:
We need this on general principle, because while ostree tries
to support both `/etc` and `/usr/etc`, in practice rpm-ostree
effectively requires ostree commits to have `/usr/etc`.

### DIFF
--- a/lib/src/tar/write.rs
+++ b/lib/src/tar/write.rs
@@ -54,6 +54,7 @@ pub async fn write_tar(
         c.args(&[
             "--no-bindings",
             "--tar-autocreate-parents",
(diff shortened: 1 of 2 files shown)


lucab:
Logic looks good, although I'm not sure at which point of the flow we ended up trying to import a tarball with content `/etc`. Are we maybe missing a validation step before reaching here, to ensure we only have content under `/usr`?

lucab (approved):

lucab:
Nevermind, https://github.com/ostreedev/ostree-rs-ext/pull/99#issuecomment-926933264 already explained how content ended up in `/etc`.

PR #109, opened 2021-09-28
tokio_util: Add API to do GLib+`GCancellable` from `async fn`

cgwalters:
This is the inverse of the previous API; we have cases today
where we're spawning a helper blocking thread that calls functions
internally that use `GCancellable`, but we weren't passed one.

See https://github.com/gtk-rs/gtk-rs-core/issues/240

### DIFF
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -219,8 +219,8 @@ impl Importer {
         size: usize,
         checksum: &str,
         xattrs: Option<glib::Variant>,
+        cancellable: Option<&gio::Cancellable>,
     ) -> Result<()> {
-        let cancellable = gio::NONE_CANCELLABLE;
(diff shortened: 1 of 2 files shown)


lib/src/tar/import.rs
```diff
@@ -606,10 +603,10 @@ pub async fn import_tar(
     let options = options.unwrap_or_default();
     let src = ReadBridge::new(src);
     let repo = repo.clone();
-    let import = tokio::task::spawn_blocking(move || {
+    let import = crate::tokio_util::spawn_blocking_cancellable(move |cancellable| {
```
lucab, line 606: Minor: a `use crate::tokio_util` in this module will make this lighter to read.

lucab (approved):

lib/src/tar/import.rs
cgwalters: Hmm, it's the only use though?

Issue #12, opened 2021-04-15
containers: support derivation


cgwalters:
We may in the future support the equivalent of a `Dockerfile` that does
```
FROM quay.io/exampleos/exampleos:latest
RUN curl ...
ADD ...
RUN ostree container-finalize-bootable
```

What this will do is basically an ostree-tar version of `ostree commit --tree=dir=/ --selinux-policy-from-base` that will do e.g. selinux labeling, hardlink deduplication and computing the sha256 digest.

In theory, we don't require this "blessing" of `ostree container-finalize-bootable` - we could do it client side too.  But I think it's better for the client side to be "dumb" by default - it makes it more like an image system.

## Use/integration with rpm-ostree and higher level projects

This project is independent of rpm-ostree and other tools.  However we need to support them; specifically e.g.:

```
FROM quay.io/coreos/fedora:stable
RUN rpm-ostree install cowsay
RUN ostree container-finalize-bootable
```

Following the above.  And so our library/model should also support what rpm-ostree natively wants to do in unpacking rpms into ostree commits for example.  

Ultimately, this scope creeps basically into a new ostree repo mode `container-tar` or so?

cgwalters:
If we want to support `Dockerfile` then it basically means we need to understand container layer deltas.

cgwalters:
Thinking through this more, it seems to me that derivation support does scope creep into mapping container image layers into ostree commits, and handling GC.  Basically to gain the real benefit, we need to avoid re-pulling layers we already have.

This isn't hard at all, and similar code exists in containers/storage's ostree backend.  But it is definitely a scope increase.

cgwalters:
OK to try this out using e.g. FCOS stable:

First, switch to a build I uploaded that has https://github.com/ostreedev/ostree-rs-ext/pull/99 plus an updated rpm-ostree, plus https://github.com/cgwalters/container-image-proxy stuck in the image too.

These instructions have been tested on current FCOS stable `34.20210904.3.0`.  But note that the syntax for rebasing to containers has changed in the newer rpm-ostree v2021.11 (with newer ostree-rs-ext) which is included in this test container I've uploaded:

```
# systemctl mask --now zincati
# rpm-ostree rebase --experimental docker://quay.io/cgwalters/fcos-dev:latest
...
# systemctl reboot
```

[...]

PR #112, opened 2021-10-01
lib/tar: Pre-filter tar archive in Rust

cgwalters:
Rather than trying to extend the ostree C code to support
arbitrary transformations, let's do pre-parsing here in safe Rust.

We keep the `/etc` → `/usr/etc` bits, but we also just completely
drop everything not in `/usr` now.

As noted in the comment, this pre-validation will hopefully also
catch any corrupt tarballs that might be exploitable in the C libarchive
codebase.

### DIFF
--- a/lib/src/async_util.rs
+++ b/lib/src/async_util.rs
@@ -1,6 +1,6 @@
 use std::io::prelude::*;
 use std::pin::Pin;
-use tokio::io::{AsyncRead, AsyncReadExt};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 /// A [`std::io::Read`] implementation backed by an asynchronous source.
 pub(crate) struct ReadBridge<T> {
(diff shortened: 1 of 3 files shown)


lib/src/tar/write.rs
```diff
@@ -49,23 +51,94 @@ fn sepolicy_from_base(repo: &ostree::Repo, base: &str) -> Result<TempSePolicy> {
...
+        let path = path
+            .strip_prefix("etc/")
+            .ok()
+            .map(|p| Cow::Owned(Utf8Path::new("usr/etc").join(p)))
+            .unwrap_or_else(|| Cow::Borrowed(path));
+        let path = &*path;
+
+        // Completely discard stuff outside of /usr
+        if !(matches!(path.as_str(), "/" | "/usr") || path.starts_with("usr/")) {
+            continue;
```
lucab, line 99: I'd throw a `warn!()` in here. I'm pretty sure we'll discover a good bunch of nasty things.

lucab (approved):

lib/src/tar/write.rs
cgwalters: Good point, though since this is a library we should probably allow the caller to collect this data and log/print it however it wants.

Will look at that.

cgwalters: In the bigger picture I am currently thinking of this as a shortcut - longer term we'll strongly encourage the model in https://github.com/ostreedev/ostree-rs-ext/issues/12#issue-859175906 where we have a required `RUN ostree container-finalize-bootable` incantation that will clean this stuff up at build time.

cgwalters:
OK, that was a fun deep dive into path libraries in Rust, specifically the [components method](https://docs.rs/camino/1.0.4/camino/struct.Utf8Path.html#method.components).

We now parse, normalize and filter the input path in a much more robust way, and this updated code now also returns a btreemap from "prefix -> number of filtered paths" which should be sufficient to start.

lib/src/async_util.rs
```diff
@@ -26,6 +26,35 @@ impl<T: AsyncRead> ReadBridge<T> {
     }
 }
 
+/// A [`std::io::Write`] implementation backed by an asynchronous source.
```
cgwalters, line 29: BTW I submitted https://github.com/tokio-rs/tokio/pull/4146

cgwalters:
Hmm, what we should really do here actually is store the metadata about filtered paths in the commit object.  However, doing that would require splitting up the write/commit phases, which in turn really wants us to be doing all of this natively in Rust instead of forking off `ostree commit`.  Which in turn depends on https://github.com/ostreedev/ostree/pull/2449 etc.  So...later.

cgwalters:
Any other comments on this, or good to go?

lucab:
All good from my side :+1:.

As (very subjective) feedback on your last comment, I don't personally feel the need to record that information persistently. It can be interesting to log and inspect it at runtime, but in the steady state I guess it's unlikely that we'd want to parse/analyze that from the commit object itself.

Issue #115, opened 2021-10-04
container: use imageid


cgwalters:
Right now we're using the manifest digest, which is useful because it's the thing one needs for pull-by-digest.

However, we can be more robust to transformations like oci -> docker schema by also using the [image id](https://github.com/opencontainers/image-spec/blob/main/config.md#imageid).  Specifically, we should also say "no changes" if the manifest has changed somehow, but the image ID is the same.

PR #116, opened 2021-10-05
container/oci: Add config to Manifest + container/oci: Use our Manifest struct when writing

cgwalters:
container/oci: Add config to Manifest

It's required, and we need it to compute the "ImageID".
xref https://github.com/ostreedev/ostree-rs-ext/issues/115

---

container/oci: Use our Manifest struct when writing

This way we get type safety, we're testing the API etc.

---

### DIFF
--- a/lib/src/container/oci.rs
+++ b/lib/src/container/oci.rs
@@ -8,10 +8,8 @@ use openat_ext::*;
 use openssl::hash::{Hasher, MessageDigest};
 use phf::phf_map;
 use serde::{Deserialize, Serialize};
-use std::{
-    collections::{BTreeMap, HashMap},
-    io::prelude::*,
-};
+use std::collections::HashMap;
+use std::io::prelude::*;
 use tokio::io::AsyncBufRead;
 
 /// Map the value from `uname -m` to the Go architecture.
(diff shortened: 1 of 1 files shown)


lib/src/container/oci.rs
```diff
@@ -230,22 +230,23 @@ impl<'a> OciWriter<'a> {
...
         let config_blob = write_json_blob(self.dir, &config)?;
 
-        let manifest_data = serde_json::json!({
-            "schemaVersion": default_schema_version(),
-            "config": {
-                "mediaType": OCI_TYPE_CONFIG_JSON,
-                "size": config_blob.size,
-                "digest": config_blob.digest_id(),
+        let mut annotations = BTreeMap::new();
+        annotations.extend(self.manifest_annotations.drain());
```
lucab, line 234: Minor: this performs an implicit sorting step, iterating through an hashmap. I think it's a symptom which highlights an impedance mismatch in `OciWriter` definition. I'd suggest to switch that to directly use a `BTreeMap` or a `Vec`, so that are annotations are already in the expected order when we reach here.

cgwalters: In practice here we use https://crates.io/crates/cjson which will order things.  But I can't imagine a case in which anything should actually depend on the ordering of the annotations.  So switched to `HashMap` for consistency.

Issue #69, opened 2021-08-27
container: support splitting inputs


cgwalters:
# Optimizing ostree-native containers into layers

This project started by creating a simple and straightforward mechanism to bidirectionally map between an ostree commit object and an OCI/Docker container - we serialize the whole commit/image into a single tarball, which is then wrapped as a container layer.

In other words, we generate a base image with one big layer.  This was simple and straightforward to implement, relatively speaking.  (Note that *derived* containers naturally do have distinct layers.  But this issue is about the "base image".)

The core problem is that *any change* to the base image (e.g. kernel security update) requires each client to redownload *the entire base image* which could be 1GB or more.  See e.g. [this spreadsheet for Fedora CoreOS content size](https://docs.google.com/spreadsheets/d/18Yl9XP_ob6sG2CvgNF15Go1l5LRYo0-f1VyPrNo-E58/edit#gid=1505373751).

# Prior art and related projects

See https://grahamc.com/blog/nix-and-layered-docker-images for a Nix based build system that knows how to intelligently generate a container image with multiple layers.

There is also work in the baseline container ecosystem for optimizing transport of containers.

Copying in bits of [this comment](https://github.com/containers/image/pull/902#issuecomment-987302590):

- https://www.balena.io/docs/learn/deploy/delta/ is a custom fork of docker with custom deltas that has been around a while
- https://github.com/containerd/stargz-snapshotter/blob/main/docs/estargz.md is a bit more recent
- https://github.com/containers/image/pull/902 is bsdiff for layers
- https://www.scrivano.org/posts/2021-10-26-compose-fs/ and [zstd chunked](https://github.com/containers/image/pull/1084)
- https://github.com/linuxkit/linuxkit changes the default OS to be a pile of containers (very interesting direction)

estargz has some design points worth looking at, but largely speaking I think few people want to have their *base operating system* lazily fetched.  (Although, there is clearly an intersection with estargz's per-file digests and ostree)

# Proposed initial solution: support for splitting layers

I think right now, we cannot *rely* on any container-level deltas.  As a baseline, we should support splitting ostree commits into distinct layers, because that works with every container runtime and registry today.  Further, splitting layers optimizes both pulls *and pushes*.

In addition, split layers means that any other "per layer" deltas actually just work more efficiently and better, whether that's zstd-chunked or layer-bsdiff.

Aside: today with Dockerfile, one cannot easily do this.  Discussion: https://github.com/containers/podman/discussions/12605

[...]

PR #123, opened 2021-10-22
Support for split layers

cgwalters:
Support for split layers

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/69

This is initial basic support for splitting files (objects) from
a commit into separate container image layers, and reassembling
those layers into a commit on the client.

We retain our present logic around e.g. GPG signature verification.

There's a new `chunking.rs` file which has logic to automatically
factor out things like the kernel/initramfs and large files.

In order to fetch these images client side, we now heavily
intermix/cross the previous code for fetching non-ostree layers.

---


cgwalters:
OK, this is finally in a state where it seems to work!  Lifting draft, but it's a huge change so we need careful review and thought.  At least the core logic today is only opt-in, you have to `ostree container encapsulate --ex-chunked ...` to enable it.

cgwalters:
Also to make use of this you need a skopeo with https://github.com/containers/skopeo/pull/1523 which isn't released yet - I pinged about that.

cgwalters:
OK now with https://github.com/ostreedev/ostree-rs-ext/pull/268 this is getting closer to ready to merge I think!

cgwalters:
This does still block on https://bodhi.fedoraproject.org/updates/FEDORA-2022-be638496e5 and...hmm, https://gitlab.com/redhat/centos-stream/rpms/skopeo/-/commit/e3b85c5431074f8ed2944f570d45803961c0a068 landed but I don't see it in C9S yet.  Oh...perhaps because https://bugzilla.redhat.com/show_bug.cgi?id=2000051 isn't marked VERIFIED or something?

cgwalters:
OK, CI on https://github.com/coreos/rpm-ostree/pull/3478 is passing!  I'd like to merge this now, I've fixed the bugs that caused it break the current v0 container format, so while it's a lot of new code, I think the risk is pretty low because most of the new bits are opt-in.

jlebon (approved):
Very cool! I've only skimmed this, but it looks sane overall.

jmarrero (approved):

PR #128, opened 2021-10-26
Use tokio SyncIoBridge

cgwalters:
We can only land this after a new tokio is released with
https://github.com/tokio-rs/tokio/pull/4146


cgwalters:
And the tokio release is done, so this should be good to go!

lucab (approved):

PR #135, opened 2021-10-28
container/deploy: Add a `--target-imgref` option

cgwalters:
We want the ability to separate the image that's pulled from its
target reference.  This is what we do today in coreos-assembler
for pure ostree refs - e.g. we deploy via a commit hash, but
still have `ostree admin upgrade` pull from a ref.

This is equivalent functionality for containers.


cgwalters:
Followup https://github.com/ostreedev/ostree-rs-ext/issues/144

PR #138, opened 2021-10-29
lib/container: Use "encapsulate" terminology for module names too

cgwalters:
Previously in https://github.com/ostreedev/ostree-rs-ext/pull/113
we changed the CLI entrypoints with this rationale:

> Since we're moving towards more "native" support for container
> images, we need to very clearly differentiate between the code
> that currently uses the terms "import" and "export" which are
> somewhat ambiguous.

To elaborate, the term "import" is ambiguous with respect to our
container "store" path, which generates a new ostree commit and
operates on arbitrary container images.

Let's take the next step now and rename the modules too.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -13,7 +13,7 @@ use std::ffi::OsString;
 use structopt::StructOpt;
 
 use crate::container::store::{LayeredImageImporter, PrepareResult};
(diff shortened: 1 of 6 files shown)


lucab (approved):

cgwalters:
(rebased :surfer: )

Issue #143, opened 2021-11-02
derivation: Support importing commit metadata, or original commit


cgwalters:
Today rpm-ostree injects a lot of commit metadata, including things like:

- hash of input state
- input rpm-md repos and their timestamps
- rpmdb
- The new `ostree.linux` key with the kernel version

Right now we toss all of that when deploying via the new layered container image flow - even if that container has just a single layer.  The reason for that is that we need to store e.g. the container image manifest.

There are a few potential directions to go:

### Map ostree metadata to oci annotations

In this we map commit metadata to [oci annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md).  A major benefit of this is things like `ostree.linux` (i.e. kernel version) would also be immediately visible in `skopeo inspect` and similar tools.

Actually this path leads towards the obvious thing of also exporting the rpmdb for container images we generate that are built via `yum`.  (Nontrivial work there)

### Store container metadata out of band

Perhaps instead of synthesizing a new commit object and dropping the original encapsulated commit on the floor, we should instead write the original commit as is, but store the manifest as e.g. local detached metadata on the commit?  Maybe use extended attributes?

Or store it as a commit, but in a new ref like `/ostree/containers/manifest/<digest>` and also retain the original commit?

cgwalters:
Splitting this out of discussion here https://github.com/coreos/coreos-assembler/pull/2523#discussion_r740539153

jlebon:
> ### Store container metadata out of band
> 
> Perhaps instead of synthesizing a new commit object and dropping the original encapsulated commit on the floor, we should instead write the original commit as is, but store the manifest as e.g. local detached metadata on the commit? Maybe use extended attributes?
> 
> Or store it as a commit, but in a new ref like `/ostree/containers/manifest/<digest>` and also retain the original commit?

Huge +1 on conserving the original commit, and deploying it too, so that `rpm-ostree status` matches. I think it'll make things much less confusing if it matches between the two worlds and preserving commit metadata and the `rpm-ostree status --json` API. Maybe we can add a `ContainerDigest:` field showing the container SHA256?

cgwalters:
The core tension here is that in the current derivation model, all the metadata we have is from the base image - unless we require a finalization step like `RUN ostree container finalize` to be run which would create a commit object.

[...]

PR #147, opened 2021-11-04
container: Rework API and model around "base commit" and "merge commit"

cgwalters:
containers: Expose a state struct with base commit and layering state

This is actually quite analogous to what rpm-ostree does internally;
a lot of the internals there distinguish "base" versus "layered"
commits, and this is very similar.

Here, when we import a non-layered image, we can mostly
ignore the merge commit since all it has is the image manifest.
The base image layer ref *is* the encapsulated ostree commit with
all the metadata injected by (rpm-)ostree in the non-layered case.

Add an API which exposes this as a struct, and also return it
from the importer's `AlreadyPresent` case.

---

container: Change import result case to contain state struct

Notably, this also stops exposing the ostree ref for the merge commit,
which I think is a good idea in general since it should be thought
of more as an implementation detail.

In other words, this module speaks container image references
and ostree commits.

---

container/deploy: Use base commit if we're not layered

If we're not doing a layered image, then use the base commit
for the deployment.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/143

---

container/deploy: Also write ref with target if provided

With this new emphasis on "dual commit objects" for the container
deployment, the higher level code queries via container image references
and we don't expose the ostree ref (since there is no longer a single
one).

This makes it critical to write the internal ref matching the
target container image, because it now needs to match the origin.

---

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -10,7 +10,7 @@ use crate::refescape;
 use anyhow::{anyhow, Context};
 use containers_image_proxy::{ImageProxy, OpenedImage};
 use fn_error_context::context;
-use oci_spec::image as oci_image;
+use oci_spec::image::{self as oci_image, ImageManifest};
 use ostree::prelude::{Cast, ToVariant};
 use ostree::{gio, glib};
 use std::collections::{BTreeMap, HashMap};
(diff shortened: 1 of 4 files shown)


lib/src/container/store.rs
```diff
@@ -355,6 +369,46 @@ pub fn list_images(repo: &ostree::Repo) -> Result<Vec<String>> {
         .collect()
 }
 
+fn query_image_impl(
+    repo: &ostree::Repo,
+    imgref: &OstreeImageReference,
+) -> Result<Option<(ImageManifest, LayeredImageState)>> {
```
cgwalters, line 397: I thought about exposing the parsed manifest in `LayeredImageState` too, but it can come later.

cgwalters:
I've updated https://github.com/coreos/rpm-ostree/pull/3139 to use this, and https://github.com/coreos/coreos-assembler/pull/2523 is also updated and I've tested them all together.

lucab (approved):

PR #149, opened 2021-11-05
container/store: Write filter data to commitmeta, clean up API

cgwalters:
The import path was previously providing the data about filtered-out
content (e.g. files in `/var`) "out of band" as part of the import
return value.  But this means that unless that data is e.g. logged
it ends up being lost.

Since the amount of data is bounded, let's instead add it as metadata
to the merge commit.

Then this lets us make a nice cleanup of entirely dropping the
`CompletedImport` struct in favor of the `LayeredImageState` that
we use in various places.

Now it's possible for something like `rpm-ostree status` to also
reliably show how many files were filtered out of each layer.

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -13,7 +13,8 @@ use fn_error_context::context;
 use oci_spec::image::{self as oci_image, ImageManifest};
 use ostree::prelude::{Cast, ToVariant};
 use ostree::{gio, glib};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
+use std::iter::FromIterator;
 
 /// The ostree ref prefix for blobs.
 const LAYER_PREFIX: &str = "ostree/container/blob";
@@ -24,6 +25,10 @@ const IMAGE_PREFIX: &str = "ostree/container/image";
 const META_MANIFEST_DIGEST: &str = "ostree.manifest-digest";
 /// The key injected into the merge commit with the manifest serialized as JSON.
 const META_MANIFEST: &str = "ostree.manifest";
+/// Value of type `a{sa{su}}` containing number of filtered out files
+pub const META_FILTERED: &str = "ostree.tar-filtered";
+/// The type used to store content filtering information with `META_FILTERED`.
+pub type MetaFilteredData = HashMap<String, HashMap<String, u32>>;
 
 /// Convert e.g. sha256:12345... into `/ostree/container/blob/sha256_2B12345...`.
 fn ref_for_blob_digest(d: &str) -> Result<String> {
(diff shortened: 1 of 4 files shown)


lib/src/container/store.rs
```diff
@@ -287,7 +282,7 @@ impl LayeredImageImporter {
         };
 
         let mut layer_commits = Vec::new();
-        let mut layer_filtered_content = BTreeMap::new();
+        let mut layer_filtered_content: MetaFilteredData = HashMap::new();
```
cgwalters, line 285: TODO: investigate supporting `BTreeMap` natively in glib-rs

lib/tests/it/main.rs
```diff
@@ -446,7 +446,7 @@ async fn test_container_write_derive() -> Result<()> {
 
     let imported_commit = &fixture
         .destrepo
-        .load_commit(import.state.merge_commit.as_str())?
+        .load_commit(import.merge_commit.as_str())?
```
cgwalters, line 449: All this code no longer needs to traverse the an intermediate struct.

lib/src/container/store.rs
cgwalters: https://github.com/gtk-rs/gtk-rs-core/pull/334

lib/src/container/store.rs
```diff
@@ -24,6 +25,10 @@ const IMAGE_PREFIX: &str = "ostree/container/image";
 const META_MANIFEST_DIGEST: &str = "ostree.manifest-digest";
 /// The key injected into the merge commit with the manifest serialized as JSON.
 const META_MANIFEST: &str = "ostree.manifest";
+/// Value of type `a{sa{su}}` containing number of filtered out files
+pub const META_FILTERED: &str = "ostree.tar-filtered";
+/// The type used to store content filtering information with `META_FILTERED`.
+pub type MetaFilteredData = HashMap<String, HashMap<String, u32>>;
```
lucab, line 31: Is this useful this way though? Wouldn't a consumer prefer a map of `layer_id` -> `set of filtered paths`?

cgwalters: This data is stored once per commit, and we have a one-one mapping of layer to commit.

cgwalters: We could add an API that loaded all the commits and printed that I guess too.

lucab: Ack, though I'm not sure what to take away from that statement. Are you saying that it would be too much data to store?

For clarity, I'm not concerned with the keys of this map, but with the values: they seem more obscure and less useful for the typical operation that a consumer would do (i.e. list/query/print entries for filtered paths).

In pseudo-types, my original comment was asking: why not doing `HashMap<LayerDigestString, Vec<FilteredPathString>>` instead?

cgwalters: Yeah, need to document this better but basically the map is e.g.:

- `/var` -> 42
- `/boot` -> 3

In order to provide some information about filtered paths, while also not having metadata that scales linearly with the content size.

lucab: > not having metadata that scales linearly with the content size

Ok thanks, that answered it.

lucab (approved):

PR #152, opened 2021-11-09
container: Add options struct for encapsulation

cgwalters:
This exposes the ability to disable compression, which is mostly
only useful for the internal flow that pushes to containers/storage
where we don't want compression.

But the main goal here is to pave the way for adding a `bool chunking`.


jlebon (approved):
Looks sane to me.

PR #156, opened 2021-11-10
Bump to containers-image-proxy 0.3, add proxy config as API and CLI

cgwalters:
Bump to containers-image-proxy 0.3

---

lib/container: Drop unnecessary `mut` for proxy

The proxy has gone back to a shared borrow with internal locking.

---

lib: Expose proxy configuration as public API

We will need to support basically all options that the proxy
does, such as disabling TLS, separate authfile etc.

Rather than write a new type, just re-export it.

Thread the new options struct through the store and deploy APIs.

---

cli: Expose current container proxy opts as CLI flags

We want this for the same reason they exist in skopeo
around non-default authfiles, etc.

The chain here is pretty amazing:

- containers/image internal Go API
- skopeo CLI flags
- containers/containers-image-proxy Rust API
- ostreedev/ostree-rs-ext Rust API
- ostreedev/ostree-rs-ext CLI code
- rpm-ostree CLI

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -10,7 +10,7 @@ version = "0.4.0"
 
 [dependencies]
 anyhow = "1.0"
-containers-image-proxy = "0.2"
+containers-image-proxy = "0.3"
 async-compression = { version = "0.3", features = ["gzip", "tokio"] }
 bytes = "1.0.1"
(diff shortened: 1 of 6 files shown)


jlebon (approved):
Looks sane to me!

Issue #162, opened 2021-11-18
fails on long symlinks


cgwalters:
```
[root@quicksilver ~]# /var/srv/walters/src/github/ostreedev/ostree-rs-ext/target/debug/ostree-ext-cli container encapsulate --repo=/ostree/repo f8e7ed0ee87a8ae48eb779c085f4fffddb824c5830e43715e6c92799a678d806 oci:/var/tmp/fedora-silverblue
error: Building oci: Writing ostree root to blob: Exporting commit: Writing content symlink: 87c8834da3bad596352375ff413bca064584f184bdd6ba1764293d137249e168: provided value is too long when setting link name for 
[root@quicksilver ~]# ostree ls -C -R f8e7ed0ee87a8ae48eb779c085f4fffddb824c5830e43715e6c92799a678d806 | grep 87c8834da3bad596352375ff413bca064584f184bdd6ba1764293d137249e168
l00777 0 0      0 87c8834da3bad596352375ff413bca064584f184bdd6ba1764293d137249e168 /usr/lib/.build-id/05/159ed904e45ff5100f7acd3d3b99fa7e27e34f -> ../../../../usr/lib64/qt5/plugins/wayland-graphics-integration-server/libqt-wayland-compositor-xcomposite-egl.so
[root@quicksilver ~]# 
```

This is https://github.com/alexcrichton/tar-rs/issues/192

In theory, we could hack around this by detecting this case and serializing to the filesystem.  Or, we could try a PR to tar-rs.

cgwalters:
PR in https://github.com/ostreedev/ostree-rs-ext/pull/165

lrodriguez14:
Hello @cgwalters, I am having the same issue using The CoreOS Assembler when adding some packages:
```
New group entries: clevis, dnsmasq, docker, fedora-coreos-pinger, kvm, printadmin, render, sgx, systemd-coredump, systemd-oom, zincati
Committing... done
Metadata Total: 20647
Metadata Written: 101
Content Total: 52637
Content Written: 5
Content Cache Hits: 20297
Content Bytes Written: 95570702
101 metadata, 291 content objects imported; 102.2 MB content written                                                 
fedora/x86_64/coreos/testing-devel => 35b514a4147b50c6f0d031da4fcd34cb7d26732c857980ace709716dd0b206ad
ostree diff commit from: ce496b15dee3e16e6ffac34db9be37e244ae3a4b9292382d734a72827ca2a376
[...]

PR #165, opened 2021-11-18
Handle long and denormal symlinks

cgwalters:
Use new `append_link()` API to handle long symlinks

I hit this when exporting Fedora Silverblue, there are some
long symlinks in there.

Depends: https://github.com/alexcrichton/tar-rs/pull/273
Closes: https://github.com/ostreedev/ostree-rs-ext/issues/162

---

tar/export: Write symlink targets literally

Requires: https://github.com/alexcrichton/tar-rs/pull/274

And I'll just copy/paste the commit message from there, lightly edited:

In https://github.com/ostreedev/ostree we generate a cryptographic
checksum over files and symlinks, and directories.

ostree does not currently perform any canonicalization on symlinks;
we'll respect and honor whatever bytes we're provided as input,
and replicate that on the target.

We're using the Rust tar crate to do tar serialization,
which has so far worked fine...except, I hit this corner case:

```
[root@cosa-devsh ~]# rpm -qf /usr/lib/systemd/systemd-sysv-install
chkconfig-1.13-2.el8.x86_64
[root@cosa-devsh ~]# ll /usr/lib/systemd/systemd-sysv-install
lrwxrwxrwx. 2 root root 24 Nov 29 18:08 /usr/lib/systemd/systemd-sysv-install -> ../../..//sbin/chkconfig
[root@cosa-devsh ~]#
```

But, using `set_link_name` to write the tarball, we end up with
the canonicalized path `../../../sbin/chkconfig` - i.e. without the
double `//`.  This breaks the checksum.

Now, I am a bit tempted to change ostree to do canonicalization.  But
even if we did, I'd need to *exactly* match what tar-rs is doing.

(I may of course also try to change the rhel8 systemd package, but
 that's going to take a while to propagate and this corner case isn't
 the only one I'm sure)

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -35,7 +35,7 @@ pin-project = "1.0"
 serde = { features = ["derive"], version = "1.0.125" }
 serde_json = "1.0.64"
 structopt = "0.3.21"
-tar = "0.4.33"
+tar = "0.4.38"
 tempfile = "3.2.0"
 tokio = { features = ["full"], version = "1" }
 tokio-util = { features = ["io-util"], version = "0.6.9" }
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -231,13 +231,11 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
                     .append_data(&mut h, &path, &mut instream)
                     .with_context(|| format!("Writing regfile {}", checksum))?;
             } else {
-                h.set_size(0);
-                h.set_entry_type(tar::EntryType::Symlink);
                 let context = || format!("Writing content symlink: {}", checksum);
-                h.set_link_name(meta.symlink_target().unwrap().as_str())
-                    .with_context(context)?;
+                h.set_entry_type(tar::EntryType::Symlink);
+                h.set_size(0);
                 self.out
-                    .append_data(&mut h, &path, &mut std::io::empty())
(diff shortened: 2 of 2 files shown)


cgwalters:
OK there's a new [tar 0.4.38](https://crates.io/crates/tar/0.4.38) with my PRs!  Because it conflicted, I'm rolling
https://github.com/ostreedev/ostree-rs-ext/pull/182
into this one.

(And now that I look at this...honestly I regret not also adding `append_link_literal`.  Oops.)

jmarrero (approved):
lgtm

JayDoubleu:
I'm getting the same error with below: 
```
rpm-ostree ex-container encapsulate --repo=repo fedora/35/x86_64/silverblue docker://ghcr.io/JayDoubleu/fedora-silverblue:35
error: Building oci: Writing ostree root to blob: Exporting commit: provided value is too long when setting link name for 
```
```
[jaydoubleu@thinkxps-sb workstation-ostree-config]$ rpm-ostree --version
rpm-ostree:
 Version: '2021.14'
 Git: 5cb5980ad94eaaee0585b8dde1a733cb12670b19
 Features:
  - compose
  - rust
  - fedora-integration
```

Does it mean I should wait for newer rpm-ostree ?

cgwalters:
Hi, yes - this fix should land in the next rpm-ostree release.  There are automated builds from git main at https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/

You can also build from source of course.

PR #167, opened 2021-11-22
A few small patches for `tar/import`

cgwalters:
tar/import: Remove unnecessary `abort_transaction()`

This is dead code since b7ba07556c8c54a719f47d9a8f1ef47b5b7a0e4b
when we switched to `auto_transaction()`.

---

tar/import: Move directory filtering into filter function

It's cleaner if our `filter_entry` function does most of the
work here instead of having half of the filtering in a closure.

---

tar/import: Move txn outside of `Importer`

This is prep for split tar imports, where we'll reuse distinct
calls to the importer across a single transaction.

---


jlebon (approved):
Looks sane to me!

PR #168, opened 2021-11-23
container/encapsulate: Write `ostree.encapsulated` annotation on layer

cgwalters:
This is prep for commit splitting into separate blobs, to help
identify layers which are "ostree native" and those which are not.

It turns out since the Docker schema:
https://docs.docker.com/registry/spec/manifest-v2-2/
does not have annotations, they simply get discarded, so we can't rely
on this.

But...I wrote this code and I think it makes sense, so let's ship it
anyways.

### DIFF
--- a/lib/src/container/ociwriter.rs
+++ b/lib/src/container/ociwriter.rs
@@ -76,7 +76,7 @@ pub(crate) struct OciWriter<'a> {
 
     cmd: Option<Vec<String>>,
 
-    layers: Vec<Layer>,
+    layers: Vec<(oci_image::Descriptor, String)>,
 }
 
 /// Write a serializable data (JSON) as an OCI blob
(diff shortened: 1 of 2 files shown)


lib/src/container/ociwriter.rs
```diff
@@ -133,7 +133,22 @@ impl<'a> OciWriter<'a> {
 
     /// Add a layer to the top of the image stack.  The firsh pushed layer becomes the root.
     pub(crate) fn push_layer(&mut self, layer: Layer) {
-        self.layers.push(layer)
+        let v: Option<HashMap<String, String>> = None;
+        self.push_layer_annotated(layer, v);
+    }
+
+    /// Add a layer to the top of the image stack.  The firsh pushed layer becomes the root.
```
jlebon, line 140: `firsh` -> `first`

Ahh, this is copy/pasted. Seems like one of the two docstrings should differ to mention annotations handling.

jlebon (approved):
Minor comment but LGTM as is too.

jlebon (approved):

PR #170, opened 2021-11-24
tar/export: Fix numeric export for objects directories

cgwalters:
Currently it looks like `sysroot/ostree/repo/objects/0xf5`
when it should obviously be
`sysroot/ostree/repo/objects/f5`.

Reported by @giuseppe


giuseppe (approved):
LGTM

PR #173, opened 2021-11-30
ci: add GH workflow for Minimum Supported Rust Version (MSRV)

lucab:
This adds a new CI jobs to make sure the code is building fine under
a given MSRV. It mostly helps making sure we stay compatible with
system toolchains shipped by distros (e.g. RHEL8).
This keeps using system dependencies from FCOS buildroot (notably,
a fresh ostree), but it drop the installed toolchain in favor
of a CI-pinned one.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,12 +14,13 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  # Minimum supported Rust version (MSRV)
+  ACTION_MSRV_TOOLCHAIN: 1.54.0
 
 jobs:
   build:
     runs-on: ubuntu-latest
     container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
-
     steps:
       - uses: actions/checkout@v2
       - name: Install deps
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

lucab:
I've never played with that, but it looks like we can place shared stuff into `ostreedev/.github` that way. I'll setup everything here first to make sure it works, then move to the shared place for ostree-rs.

lucab:
The MSRV flow seems to work. It is green on current `main`, and stacking a revert of https://github.com/ostreedev/ostree-rs-ext/pull/172 on top of that turns it red.

cgwalters (approved):

lucab:
I had a look at the suggested workflow templates, and my current recommendation is not go with them.
The reasons are:
 * Those are orga-wide templates. They still need to be instantiated/committed to each repo separetely, see the last section of https://github.blog/2020-06-22-promote-consistency-across-your-organization-with-workflow-templates/
 * They do not increase centralization. They don't even help with de-duplication as in fact they add another (templated) copy of the workflow. The template and each instance needs to be maintained separately.
 * There seems to be no automatic/assisted mechanism for propagating template updates to each workflow instance. Also, the downstream instance can be freely changed directly and get out of sync with the parent template.
 * There seems to be no automatic gating of template changes against the downstream consumers, effectively allowing breaking+gating changes to go unnoticed until after merged and manually propagated.

Looking around, I think what we actually want is this requested feature for reusable workflows: https://github.com/github/roadmap/issues/98. Even then though, there are still open problems about how to test changes at the consuming sites and how to automatically propagate updates.

PR #175, opened 2021-12-01
ci: add GH workflow for linting 

lucab:
This fixes the following warnings highlighted by clippy:
 * https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into
 * https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

It also adds a new CI job dedicated to linting, which includes steps for
rustfmt and clippy.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,6 +16,8 @@ env:
   CARGO_TERM_COLOR: always
   # Minimum supported Rust version (MSRV)
   ACTION_MSRV_TOOLCHAIN: 1.54.0
+  # Pinned toolchain for linting
+  ACTION_LINTS_TOOLCHAIN: 1.56.0
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


lib/src/container/store.rs
```diff
@@ -487,3 +487,13 @@ pub fn prune_images(_repo: &ostree::Repo, _imgs: &[&str]) -> Result<()> {
     // And remove the unreachable ones.
     unimplemented!()
 }
+
+impl From<cli::ContainerProxyOpts> for ImageProxyConfig {
```
cgwalters, line 491: While this is just one crate today, this feels like a layering violation to have the `lib/` code import `cli/`.

IOW I feel the clippy lint is correct in general, but in this specific case using `Into` is better, and we can just annotate with `#[allow(clippy::from_over_into)]`.  
Do you agree?

lucab: No strong feelings on either approaches, I can push it in that direction.

But just to double-check, are you perhaps misquoting this? Both the source and target types here are in the same `lib/` crate, we never cross to the `cli/` one.

cgwalters: Right, today the CLI code is actually all in `lib/` so that rpm-ostree can re-export it.

But the principle still stands I think.  

(Also we could fix this by keeping the `From<>` implementation in `cli.rs` right?)

lucab: Ack. Then I'll happily revert my change and just silence clippy on this one. The previous code arrangement is fine for me.

cgwalters (approved):

PR #176, opened 2021-12-06
container/store: Also parse and save image configuration

cgwalters:
Prep for using this for copies.


cgwalters:
Depends https://github.com/containers/containers-image-proxy-rs/pull/21

cgwalters:
OK now that we're tracking proxy git main, this should work.

jmarrero (approved):
lgtm

PR #178, opened 2021-12-07
Add some missing `#[derive(Debug)]`

cgwalters:
Add some missing `#[derive(Debug)]`

Hit this when I wanted to add `dbg!`.

---

Drop ununused `bytes` dependency

We aren't doing HTTP directly anymore.

---

lib: Also add `unreachable_pub` to opted-in lint

And link to the list.

I plan to cargo cult this elsewhere.

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -12,7 +12,6 @@ version = "0.5.1"
 anyhow = "1.0"
 containers-image-proxy = "0.3"
 async-compression = { version = "0.3", features = ["gzip", "tokio"] }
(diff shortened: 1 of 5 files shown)


lucab (approved):
LGTM! Would we be able to turn on the deny-lint for this, or are we missing too many on them (e.g. from dependency)?

jlebon (approved):

cgwalters:
> Would we be able to turn on the deny-lint for this

Ohhh nice.  Didn't realize that existed.  Updated :new:

Issue #121, opened 2021-10-20
design clean support for "image pull inside kubernetes"


cgwalters:
Breaking this issue out from discussion in https://github.com/openshift/os/pull/657

We need to make it easy for a Kubernetes node to pull container images for host OS updates in the same way kubelet is doing so.  In particular, I want to support pulling from e.g. an in-cluster registry.

There are *three* major aspects to this:

- Configuring pull secrets; this can be done today by injecting `REGISTRY_AUTH_FILE` in the environment.  We should also add CLI flags that just wrap the global skopeo `--authfile` argument.
- Ensuring that we trust [the service CA](https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html#understanding-service-serving_service-serving-certificate)
- Joining the pod network - we need to use cluster DNS at least.  Digging into things, it seems at least crio has special support for forking a copy of its binary into a designated cgroup.  Presumably this aids applying separate cpu/network/io limits to image pulls.

cgwalters:
My strawman would be that we expose in this library something like:

```
pub struct ImagePullConfiguration {
  pub auth_file: Option<String>,
  pub network_config: ???,
  pub cgroup: Option<String>,
}
```
or so, and then implement it by extending how we fork skopeo.

While we're here, I actually also really want to support `systemd-run skopeo` when run as root so we can use e.g. `-p DynamicUser=yes` to completely drop privileges.

cgwalters:
containers-image-proxy now has [a config struct](https://docs.rs/containers-image-proxy/0.3.1/containers_image_proxy/struct.ImageProxyConfig.html).

It's still missing the certificate bits though, and that only allows us to configure for pulls from inside a cluster, it doesn't actually demonstrate it working.

mkenigs:
Trying to understand the conclusion about networking from https://github.com/openshift/os/pull/657 - will we need to do anything special for networking? After debugging a node and oc logging in I can run `skopeo inspect docker://image-registry.openshift-image-registry.svc:5000/openshift/cli --username kubeadmin --password $(oc whoami -t)` So I'm assuming it'd also work with just `--authfile`?

Or do we need to fork into a cgroup for resource limits?

mkenigs:
Would we want to copy or factor out somewhere the `systemd-run` code from https://github.com/coreos/rpm-ostree/blob/main/rust/src/isolation.rs?

cgwalters:
[...]

PR #180, opened 2021-12-09
cli: Expose certificate_directory

mkenigs:
Exposes the corresponding options from containers-image-proxy and
skopeo

Also changes authfile type from String to PathBuf for consistency

Helps https://github.com/ostreedev/ostree-rs-ext/issues/121
Depends https://github.com/containers/containers-image-proxy-rs/pull/22

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -10,6 +10,7 @@ use ostree::{gio, glib};
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::ffi::OsString;
+use std::path::PathBuf;
 use structopt::StructOpt;
 
 use crate::container as ostree_container;
@@ -130,7 +131,12 @@ enum ContainerOpts {
 struct ContainerProxyOpts {
     #[structopt(long)]
     /// Path to Docker-formatted authentication file.
-    authfile: Option<String>,
+    authfile: Option<PathBuf>,
+
+    #[structopt(long)]
+    /// Directory with certificates (*.crt, *.cert, *.key) used to connect to registry
+    /// Equivalent to `skopeo --cert-dir`
+    cert_dir: Option<PathBuf>,
 
     #[structopt(long)]
     /// Skip TLS verification.
@@ -243,6 +249,7 @@ impl Into<ostree_container::store::ImageProxyConfig> for ContainerProxyOpts {
     fn into(self) -> ostree_container::store::ImageProxyConfig {
         ostree_container::store::ImageProxyConfig {
             authfile: self.authfile,
+            certificate_directory: self.cert_dir,
             insecure_skip_tls_verification: Some(self.insecure_skip_tls_verification),
             ..Default::default()
         }
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -235,6 +235,7 @@ struct ImaSignOpts {
 #[derive(Debug, StructOpt)]
 #[structopt(name = "ostree-ext")]
 #[structopt(rename_all = "kebab-case")]
+#[allow(clippy::large_enum_variant)]
 enum Opt {
     /// Import and export to tar
     Tar(TarOpts),
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -10,7 +10,9 @@ version = "0.5.1"
 
 [dependencies]
 anyhow = "1.0"
-containers-image-proxy = "0.3"
+# containers-image-proxy = "0.3"
+containers-image-proxy = { git = "https://github.com/containers/containers-image-proxy-rs" }
+
 async-compression = { version = "0.3", features = ["gzip", "tokio"] }
 bitflags = "1"
 camino = "1.0.4"


mkenigs:
@cgwalters are we going to have defaults/a config file for `cert_dir`? If so, what? What about `authfile`?

How do I test this with `rpm-ostree`? I've patched the build. Or do I just test `ostree-ext-cli container image pull` inside a cluster?

lib/src/cli.rs
```diff
@@ -130,7 +131,12 @@ enum ContainerOpts {
 struct ContainerProxyOpts {
     #[structopt(long)]
     /// Path to Docker-formatted authentication file.
-    authfile: Option<String>,
+    authfile: Option<PathBuf>,
+
+    #[structopt(long)]
+    /// Directory with certificates (*.crt, *.cert, *.key) used to connect to registry
+    /// Equivalent to `skopeo --cert-dir`
```
mkenigs, line 138: Is this too long for the help text?

cgwalters:
> @cgwalters are we going to have defaults/a config file for cert_dir? If so, what? What about authfile?

I think we should.  There's some bike-shedding involved for that.  My strawman would be supporting files in `/etc/ostree`.  On detail: I think we should support these being symbolic links, so that e.g. for OpenShift/MCO we can just `ln -s /var/lib/kubelet/config.json /etc/ostree/container-authfile.json` or so.

(Now the kubelet pull secret should really be in `/etc`; that's https://github.com/openshift/machine-config-operator/issues/1190#issuecomment-544246192 )

lib/src/cli.rs
cgwalters: Hmm.  I just tried this out and it looks like structopt eats newlines, but does do auto-wrapping.  I just tried e.g.:

```
diff --git a/lib/src/cli.rs b/lib/src/cli.rs
index e6cbaee..9726fcb 100644
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -162,6 +162,7 @@ enum ContainerImageOpts {
     },
 
     /// Copy a pulled container image from one repo to another.
+    /// Blah blah.  Lorem ipsum foo bar baz. Lorem ipsum foo bar baz. Lorem ipsum foo bar baz. Lorem ipsum foo bar baz.
     Copy {
         /// Path to the source repository
         #[structopt(long)]
```

And I get:

```
walters@toolbox /v/s/w/s/g/o/ostree-rs-ext (main)> cargo b; and ./target/debug/ostree-ext-cli container image --help
   Compiling ostree-ext v0.5.1 (/var/srv/walters/src/github/ostreedev/ostree-rs-ext/lib)
   Compiling ostree-ext-cli v0.1.4 (/var/srv/walters/src/github/ostreedev/ostree-rs-ext/cli)
    Finished dev [optimized + debuginfo] target(s) in 6.94s
ostree-ext-cli-container-image 0.5.1
Commands for working with (possibly layered, non-encapsulated) container images

USAGE:
    ostree-ext-cli container image <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    copy      Copy a pulled container image from one repo to another. Blah blah.  Lorem ipsum foo bar baz. Lorem
              ipsum foo bar baz. Lorem ipsum foo bar baz. Lorem ipsum foo bar baz
    deploy    Perform initial deployment for a container image
    help      Prints this message or the help of the given subcommand(s)
    list      List container images
    pull      Pull (or update) a container image
walters@toolbox /v/s/w/s/g/o/ostree-rs-ext (main)> 
```

Soo...it should be fine if you just add a period `.` at the end of sentences.  Or it could be:
```
/// Directory with certificates (*.crt, *.cert, *.key) used to connect to registry; equivalent to `skopeo --cert-dir`.
```
or so?

cgwalters:
@lucab there's a crate for systemd-style config right?  I vaguely recall something in coreos using it; tried looking in afterburn but didn't see it.

cgwalters:
> How do I test this with rpm-ostree? I've patched the build. Or do I just test ostree-ext-cli container image pull inside a cluster?

Yep, the latter should work fine!  One trick here I've used is `oc cp` to copy binaries built from my local workstation to a remote cluster.

cgwalters:
OK so to make the CI here green, we need to change our `Cargo.toml` to pull from git.  This is another variant of https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section

Do you want to do that or OK if I just push the fixups for this?

cgwalters:
@mkenigs I put together https://gist.github.com/cgwalters/ce13e9d936ea2dd55ad9f159c2e99f68

lib/src/cli.rs
mkenigs: Never added the period but not sure that's important

mkenigs:
> @lucab there's a crate for systemd-style config right? I vaguely recall something in coreos using it; tried looking in afterburn but didn't see it.

@lucab did you ever see that?

PR #183, opened 2021-12-10
container: Add history struct to oci writing, tweak history description

cgwalters:
Prep for https://github.com/ostreedev/ostree-rs-ext/issues/69
where we'll split up the input ostree commit into content-addressed
blobs.

We want to inject something useful into the the `history` in the
config git that describes each chunk, so add support for that
into our OCI writer.

Change the default description for the (currently single) layer
to include the commit subject, if present; otherwise the
commit hash.

The description of the layer shouldn't change as this tool changes,
just as the input changes.

(Side note; today rpm-ostree isn't adding a subject description,
 but hey, maybe someone else is)


jmarrero (approved):
/lgtm

cgwalters:
Thanks for the review!

PR #186, opened 2021-12-15
tar: Write directory hierarchy correctly

cgwalters:
I hit on the fact that we were missing `/tmp` in the exported
container, and this is because we were actually missing writing
directory entries.  It only works because container runtimes
will auto-create parent directories.

But we clearly want to reflect the intended uid/gid/mode in the
tar stream too, and we definitely want empty toplevel dirs like
`/tmp`.

The simple fix of emitting them in our current flow actually fails
when trying to import into `containers/storage`, complaining about
duplicate entries.

And this is because the simple fix ends up writing a `sysroot` entry
with two different modes (0755 and 0700).  One or the other needs
to win, let's just have it be 0755 for now, though it doesn't really
matter for this.

So to really make this work, rework the flow so that the tar stream
looks like:

- root directory
- sysroot/ base structure
- commit object
- commit metadata
- contents of root recursively *except* sysroot

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -30,7 +30,8 @@ oci-spec = "0.5.0"
 openat = "0.1.20"
 openat-ext = "0.2.0"
 openssl = "0.10.33"
-ostree = { features = ["v2021_5"], version = "0.13.3" }
+# ostree = { features = ["v2021_5"], version = "0.13.3" }
+ostree = { git = "https://github.com/ostreedev/ostree-rs", features = ["v2021_5"] }
 phf = { features = ["macros"], version = "0.10" }
 pin-project = "1.0"
 serde = { features = ["derive"], version = "1.0.125" }
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -14,6 +14,8 @@ use std::borrow::Cow;
 use std::collections::HashSet;
 use std::io::BufReader;
 
+// This is both special in the tar stream *and* it's in the ostree commit.
+const SYSROOT: &str = "sysroot";
 // This way the default ostree -> sysroot/ostree symlink works.
 const OSTREEDIR: &str = "sysroot/ostree";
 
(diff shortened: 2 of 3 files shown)


cgwalters:
I also did https://github.com/ostreedev/ostree-rs/pull/42 - we could make this depend on it too.

cgwalters:
Yeah, let's hold this and I'll update it to use ostree git once the above merges.

cgwalters:
Ooh actually, this breaks container copying, looking:

```
FATA[0004] writing blob: adding layer with blob "sha256:0c0adf86bf80ea77af4128ec7583695cd442fb3085e78a8fc36fbf60497c3484": Error processing tar file(duplicates of file paths not supported):  
```

Clearly we need to have real e2e tests in this repo...

cgwalters:
OK, that was fun to debug.  But it works now.  The way I was testing is:

```
$ skopeo copy oci-archive:builds/latest/x86_64/fedora-coreos-35.20211215.dev.5-ostree.x86_64.ociarchive containers-storage:localhost/fcos
Getting image source signatures
Copying blob 7803adb9de70 done  
FATA[0004] writing blob: adding layer with blob "sha256:7803adb9de70565847cc3b8c3353cc44f8b4cc39b25b8e872785048f790d6972": Error processing tar file(duplicates of file paths not supported):  
ERRO[0004] exit status 1
```

Which, okay that error would be way more useful if it told me *which* path, so I did:
```
diff --git a/vendor/github.com/vbatts/tar-split/tar/storage/packer.go b/vendor/github.com/vbatts/tar-split/tar/storage/packer.go
index aba69481..20f782e2 100644
--- a/vendor/github.com/vbatts/tar-split/tar/storage/packer.go
+++ b/vendor/github.com/vbatts/tar-split/tar/storage/packer.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"path/filepath"
 	"unicode/utf8"
@@ -47,7 +48,7 @@ func (jup *jsonUnpacker) Next() (*Entry, error) {
 	if e.Type == FileType {
 		cName := filepath.Clean(e.GetName())
 		if _, ok := jup.seen[cName]; ok {
-			return nil, ErrDuplicatePath
+			return nil, fmt.Errorf("path %s: %v", cName, ErrDuplicatePath)
 		}
 		jup.seen[cName] = struct{}{}
 	}
@@ -88,7 +89,8 @@ func (jp *jsonPacker) AddEntry(e Entry) (int, error) {
 	if e.Type == FileType {
 		cName := filepath.Clean(e.GetName())
 		if _, ok := jp.seen[cName]; ok {
-			return -1, ErrDuplicatePath
+			return -1, fmt.Errorf("path %s: %v", cName, ErrDuplicatePath)
+
 		}
 		jp.seen[cName] = struct{}{}
 	}
```

And presto:
```
$ skopeo copy oci-archive:builds/latest/x86_64/fedora-coreos-35.20211215.dev.5-ostree.x86_64.ociarchive containers-storage:localhost/fcos
Getting image source signatures
Copying blob 7803adb9de70 done  
FATA[0004] writing blob: adding layer with blob "sha256:7803adb9de70565847cc3b8c3353cc44f8b4cc39b25b8e872785048f790d6972": Error processing tar file(path sysroot: duplicates of file paths not supported):  
ERRO[0004] exit status 1                   
```

cgwalters:
OK, this should be good to go!

jmarrero (approved):
lgtm

PR #187, opened 2021-12-16
cli: Use `split_once` to parse `key=value`

cgwalters:
Motivated by dropping an `unwrap()` call.
This is the example in the doc even!

I remembered this after seeing
https://github.com/coreos/rpm-ostree/pull/3281


jmarrero (approved):
this is pretty cool syntax. lgtm

Issue #159, opened 2021-11-16
Add (and possibly require) `ostree container commit`


cgwalters:
The current layered container code performs "ostree importing" client side - we accept arbitrary layer tarballs that could come from a `Dockerfile` or other container build system.

Let's consider adding a command line `ostree container container` that is required to be invoked
in the container at the end of a build process.  For example in a `Dockerfile`:

```
FROM quay.io/fedora/coreos
ADD somefile.conf /etc/systemd/system
RUN curl ... && yum install .. && ostree container commit
```

The last step would effectively do most of the heavy lifting we're already doing client side:

- Move /etc -> /usr/etc (with backcompat link)
- Discard content outside of /usr - or better, *error out*
- Perform selinux labeling
- Compute ostree sha256

In this, I think the biggest benefit is that we can error out if people try to do things like ship content in `/boot` or `/var` at build time.  But avoiding doing selinux labeling on the client is a not-small benefit too.

Note that in order for this to work, it must be invoked *for each layer*.

cgwalters:
This issue also relates to https://github.com/ostreedev/ostree/issues/2480 so the CLI can just be `ostree` instead of `ostree-ext-cli`.

lucab:
Below some notes from an out-of-band chat that I forgot to write down earlier:
 * this sounds like a subset of what `rpm-ostree compose postprocess` does. There is possibly some logic which could be shared between the two.
 * rpm-ostree does automatic translation of /var into tmpfiles.d fragment too. But I'm unsure whether the logic here does assume systemd-tmpfiles (and other systemd mechanisms in general).
 *  depending on the starting image, it may not readily be in a runnable state. For example, FCOS images require some inflating steps (for processing tmpfiles.d, sysusers.d, and more). This in-container flow may need a symmetrical `prepare`/`finalize` logical scope to take care of that.

cgwalters:
One thing that would likely ease implementation of this significantly is if we structure it as:

- First enhance the https://docs.rs/ostree-ext/latest/ostree_ext/tar/fn.import_tar.html API to support loading selinux policy from the running root i.e. `/` instead of requiring a base commit
- generate tarball stream from `/usr` in the running container using the https://docs.rs/tar/ API
- This would include things like transforming `etc` into `usr/etc` OR we could first mutate the running root
- pipe that tar stream to our own tar import API

And that may be it!

cgwalters:
[...]

PR #189, opened 2021-12-16
Add new integration testing flow

cgwalters:
The core idea here is that we take our built binary and inject
it into a fcos container image, where we can do further testing.

For now just to prove this out I've added an internal testutils
command.

(I think I'd like to write more tests in Rust, but to do that
 correctly we also want a `bin-unit-tests` feature like rpm-ostree
 has, which can come later)

### DIFF
--- a/lib/src/integrationtest.rs
+++ b/lib/src/integrationtest.rs
@@ -1,0 +1,14 @@
+//! Module used for integration tests; should not be public.
+
+fn has_ostree() -> bool {
+    std::path::Path::new("/sysroot/ostree/repo").exists()
+}
+
+pub(crate) fn detectenv() -> &'static str {
+    match (crate::container_utils::running_in_container(), has_ostree()) {
+        (true, true) => "ostree-container",
+        (true, false) => "container",
+        (false, true) => "ostree",
+        (false, false) => "none",
+    }
+}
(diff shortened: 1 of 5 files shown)


lib/src/integrationtest.rs
```diff
@@ -0,0 +1,26 @@
+//! Module used for integration tests; should not be public.
+
+use std::path::Path;
+
+fn is_container() -> bool {
+    // https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker
+    for p in ["/run/.containerenv", "/.dockerenv"] {
```
cgwalters, line 7: OK this was fun.  I wish I could add a comment on https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker but apparently I don't have enough reputation.

cgwalters:
The thing where GH actions gives you a full VM but makes it trivial to run in a (potentially privileged) container there, or jump between containers is actually fairly powerful.

This is proving out "integration tests" that jump from a build environment into another container.

The main thought here is that with this we can start on https://github.com/ostreedev/ostree-rs-ext/issues/159 and because we'll now need to be testing ostree-rs-ext inside a container.

That said, we clearly also want VM tests.

jmarrero (approved):
lgtm

PR #190, opened 2021-12-16
Add a container_utils module with API to detect containerization

cgwalters:
This isn't going to be perfect, but it will serve for now.

We will need it in both this codebase as well as rpm-ostree, so let's
make it a public API.

### DIFF
--- a/lib/src/container_utils.rs
+++ b/lib/src/container_utils.rs
@@ -1,0 +1,17 @@
+//! Helpers for interacting with containers at runtime.
+
+/// Attempts to detect if the current process is running inside a container.
+/// This looks for the `container` environment variable or the presence
+/// of Docker or podman's more generic `/run/.containerenv`.
+pub fn running_in_container() -> bool {
+    if std::env::var_os("container").is_some() {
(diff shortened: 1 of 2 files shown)


cgwalters:
@jmarrero This relates to what you're working on too, once this lands it may be worth updating your PR to pull from git main here.

jmarrero (approved):
thanks for adding this.

cgwalters:
We had a chat in a private proprietary chat system about this, and one thing that came up is https://systemd.io/CONTAINER_INTERFACE

Which we can't use because in the default podman/docker model we *are* pid 1, and in some cases it's actually not allowed to look at `/proc/1/environ` - for example, when one is in user namespace, but with the host pid namespace (default toolbox model).

We should eventually handle the systemd-in-container path, but it has implications for ostree which is all about managing the kernel.  And to do that, we should be detecting if we're under systemd with e.g. https://docs.rs/systemd/0.10.0/systemd/daemon/fn.booted.html - and then we may need to detect "outer containerization" the systemd suggested way.

PR #191, opened 2021-12-17
tar/export: Do emit dirmeta for `/sysroot`

cgwalters:
This fixes an important regression from https://github.com/ostreedev/ostree-rs-ext/pull/186

Basically we have now a really ugly hack around `/sysroot` that
wasn't caught by the unit tests here because the OS content doesn't
have it.  I'm going to do a followup where we extend our integration
testing more here.

But anyways, we still need to emit the dirmeta object for `/sysroot`,
even if not its contents.  (Actually, it *should* be empty in an
ostree commit, hmm probably ostree should enforce that)


cgwalters:
xref https://github.com/coreos/rpm-ostree/pull/3280#issuecomment-996781067

PR #192, opened 2021-12-17
tar/export: Do checksum → string a bit smarter

cgwalters:
I was working on this area of code for a previous PR, and I almost
made a refactoring that shared the `hexbuf`.  I think the stack
allocation here was a premature optimization.  Let's just heap
allocate - modern allocators are fast with thread-local
storage and caches etc.  The C side allocates a ton anyways.

However there is an optimization we can make here - by passing
ownership of the checksum string into the recursive dirtree walk,
we can drop it before recursing.   Now there's only one heap
allocation there instead of it being O(depth).


cgwalters:
I also did https://github.com/KokaKiwi/rust-hex/pull/66

jmarrero (approved):
lgtm

PR #193, opened 2021-12-17
tar/export: Add a repo/config with mode `bare-split-xattrs`

cgwalters:
tar/export: Add a repo/config with mode `bare-split-xattrs`

Part of https://github.com/ostreedev/ostree/issues/2499

Basically, ostree core should have first-class support for
reading and parsing this.  The primary motivation is to support
ostree-in-container better.

The real main wrinkle here is xattrs, as usual.  We can't rely on
being able to read/write xattrs inside a container.

For now, this will serve sufficient to *identify* this format,
but it will pave the way for ostree core to support read/write.

(Note that if we go to implement writes, a notable wrinkle is that
 garbage collection of the `xattrs` subdirectory will require
 another special handling pass)

---

container_utils: Add a method to detect `bare-split-xattrs`

This will be used for an integration test once we rebuild
our current container image.

But also, this will help out cases like
https://github.com/coreos/rpm-ostree/pull/3280

As well as supporting finalization inside a container.

---


jmarrero (commented):
looks like linter is not happy, otherwise lgtm

jmarrero (approved):
lgtm, thanks I see the api to do the ostree check.

PR #196, opened 2021-12-21
tar: Move repo config into `/sysroot/config` by default for backcompat

cgwalters:
It turns out adding `/sysroot/ostree/repo/config` into the export
stream broke compat with older code.  Now in theory this is all
experimental and changeable, but it breaks the FCOS upgrade testing
because the export format there is "production" for some values of
production.

And for our own sanity, it's helpful to be compatible.
And add format versioning into the tar export code, and default to v0
where we put the config into `/sysroot/config` where it will be ignored
by older ostree.

This in particular should allow us to update rpm-ostree to the
latest ostree-ext in https://github.com/coreos/rpm-ostree/pull/3285

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -55,6 +55,10 @@ struct ExportOpts {
     #[structopt(long)]
     repo: String,
 
+    /// The format version.  Must be 0 or 1.
+    #[structopt(long)]
+    format_version: u32,
+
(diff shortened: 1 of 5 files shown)


jmarrero (approved):
this seems a lot more of work than what I expected. Thank you for doing this. lgtm

PR #197, opened 2021-12-22
lib/container_utils: Add helper `is_ostree_contaner()`

cgwalters:
This came up in https://github.com/coreos/rpm-ostree/pull/3297

Most of our checks are going to be for this.

### DIFF
--- a/lib/src/container_utils.rs
+++ b/lib/src/container_utils.rs
@@ -61,3 +61,10 @@ pub fn is_bare_split_xattrs() -> Result<bool> {
         Ok(false)
     }
 }
+
+/// Returns `true` if the current booted filesystem appears to be an ostree-native container.
+///
+/// This just invokes [`is_bare_split_xattrs`] and [`running_in_container`].
+pub fn is_ostree_container() -> Result<bool> {
+    Ok(running_in_container() && is_bare_split_xattrs()?)
+}


jmarrero (approved):
lgtm

lib/src/container_utils.rs
```diff
@@ -61,3 +61,10 @@ pub fn is_bare_split_xattrs() -> Result<bool> {
         Ok(false)
     }
 }
+
+/// Returns `true` if the current booted filesystem appears to be an ostree-native container.
+///
+/// This just invokes [`is_bare_split_xattrs`] and [`running_in_container`].
+pub fn is_ostree_container() -> Result<bool> {
```
lucab, line 68: Looking at this from the consumer side, how should a caller react to getting an `Err` here?
In other words, should this maybe directly map the currently "unknown" (Err) case to a plain "no" (false)?

cgwalters: I thought about that a bit.  Currently, we'll error out on various things such as I/O errors.  There's also the case of failing to parse the config file.

I feel like those should all be fatal and not silently ignored.

Or to flip it around, what errors do you think should be silently ignored?

(BTW this is related to https://users.rust-lang.org/t/why-i-use-anyhow-error-even-in-libraries/68592 )

cgwalters: (Also, sorry I didn't realize you were reviewing too, I clicked the merge button before I saw this)

lucab: No problem for the mid-fly collision, it wasn't a blocking comment.
It just feels a bit odd because we do ignore I/O errors already on half of this, as `running_in_container` uses `Path::exists` which internally maps to `metadata().is_ok()`. Which doesn't look too bad to me, assuming we mainly care about real-`true`s and can tolerate fuzzy-`false`s.

The two existing consumers in rpm-ostree would be fine both ways I think. If you had thoughts and decided in favor of explicit errors, I'm on board with that :+1:

cgwalters: https://internals.rust-lang.org/t/the-api-of-path-exists-encourages-broken-code/13817

PR #198, opened 2022-01-04
Generalize `ocidir`, dynamically generate containers

cgwalters:
Currently we have a hardcoded "exampleos.tar" fixture, as well as
two `.ociarchive` containers that also hardcode its content.

Having pre-generated container fixtures on one hand is good because
we're testing compatibility of import from a frozen format.  But it also
means that we can't easily test changing our output format.

And I'd like to introduce some changes to the output format.  I'll
reintroduce compat testing another way.

To accomplish this, rework our internal `ociwriter` to be a generalized
`ocidir` module that also supports reading (and modifying) an existing
directory.

The export/encapsulation code uses this.

Then, add a helper method that is only exposed to our integration
testing suite which uses this to generate a derived image dynamically.


cgwalters:
Draft since this depends on https://github.com/containers/oci-spec-rs/pull/85

cgwalters:
OK, latest oci-spec is released and we're now using the crate version again; this should be good to go.

cgwalters:
Rebased :surf

jmarrero:
time to rebase again :) and looks like the linter is not happy.

cgwalters:
Hmm actually it may be better if this one waits till the next release, it's not critical.

jmarrero (approved):
lgtm

PR #201, opened 2022-01-04
Use rustix instead of nix

cgwalters:
In this actually, the owned/borrowed FD stuff is definitely nicer.  I still need to investigate whether the rustix ecosystem has something like this fd passing-to-`Command` that we need.


cgwalters:
Depends https://github.com/ostreedev/ostree-rs-ext/pull/200 which is a clean prep patch that can be merged now.

cgwalters:
OK lifting draft on this one.  This also relates to https://github.com/coreos/rpm-ostree/pull/3311 in that we want to drop nix/openat from the ostree dependency chain ideally first.

cgwalters:
(We should use cap-std here too, but that's a larger churn)

jmarrero (approved):
lgtm

jmarrero:
Changes lgtm, I wonder why CI is failing on both PRs, do we need to force these in?

cgwalters:
I clearly should have tested https://github.com/ostreedev/ostree-rs-ext/pull/203 locally - we need that first because we're pulling containers-image-proxy from git now, so CI started failing after https://github.com/containers/containers-image-proxy-rs/pull/24 was merged.

(I did actually test that PR but with a different branch)

jmarrero:
I guess we need a rebase now.

cgwalters:
Rebased :surfer:

PR #202, opened 2022-01-04
lib/src/cli: Update BuildOpts fields to denote they might not be read

jmarrero:
This fixes the above warning:
```
cargo build
   Compiling ostree-ext v0.5.1 (/var/home/jmarrero/Development/github/ostreedev/ostree-rs-ext/lib)
warning: field is never read: `repo`
  --> lib/src/cli.rs:31:5
   |
31 |     repo: String,
   |     ^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `ostree_ref`
  --> lib/src/cli.rs:34:5
   |
34 |     ostree_ref: String,
   |     ^^^^^^^^^^^^^^^^^^

warning: field is never read: `oci_dir`
  --> lib/src/cli.rs:37:5
   |
37 |     oci_dir: String,
   |     ^^^^^^^^^^^^^^^

warning: `ostree-ext` (lib) generated 3 warnings
```

Also making the struct and fields pub will fix it but since the other struct in the module are not pub I am just adding the underscore here.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -25,18 +25,6 @@ fn parse_base_imgref(s: &str) -> Result<ImageReference> {
     ImageReference::try_from(s)
 }
 
-#[derive(Debug, StructOpt)]
-struct BuildOpts {
-    #[structopt(long)]
-    repo: String,
-
-    #[structopt(long = "ref")]
-    ostree_ref: String,
-
-    #[structopt(long)]
-    oci_dir: String,
-}
-
 /// Options for importing a tar archive.
 #[derive(Debug, StructOpt)]
 struct ImportOpts {


cgwalters:
Hmm, as far as I can tell, `BuildOpts` is unused.  Annotate points to https://github.com/ostreedev/ostree-rs-ext/commit/7a5cac910ca90d31c55bff5aed5c20921d4e1d31 but that's just code motion from https://github.com/ostreedev/ostree-rs-ext/commit/e011ce29206964b7e2f1a2caa0c973951eb72ded

And...well, I don't think it's worth digging farther.  Let's just delete the struct entirely?

cgwalters:
IOW I think this is the new rust compiler warning about unused fields correctly catching dead code.  What isn't clear at all to me is why it isn't saying the whole struct is unused.  OK, it has something to do with the derives, because e.g.

```
diff --git a/lib/src/cli.rs b/lib/src/cli.rs
index c96061d..48020c9 100644
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -25,15 +25,11 @@ fn parse_base_imgref(s: &str) -> Result<ImageReference> {
     ImageReference::try_from(s)
 }
 
-#[derive(Debug, StructOpt)]
 struct BuildOpts {
-    #[structopt(long)]
     repo: String,
 
-    #[structopt(long = "ref")]
     ostree_ref: String,
 
-    #[structopt(long)]
     oci_dir: String,
 }
 
```
Then gets me

```
struct is never constructed: `BuildOpts`
`#[warn(dead_code)]` on by default
```
as expected.

Anyways let's just assume the struct is some leftover relic from my early hacking days and remove it entirely.

jmarrero:
> 
> Anyways let's just assume the struct is some leftover relic from my early hacking days and remove it entirely.

Done, thanks for explaining the warning + derives relationship

cgwalters (approved):

PR #203, opened 2022-01-05
lib: Switch to released containers-image-proxy version

cgwalters:
We cut a new version upstream, let's switch back to crate dependencies.


jmarrero (approved):
lgtm

cgwalters:
Oops right, we actually need to do some porting here

PR #205, opened 2022-01-07
lib/src/cli: Add commit verb that validates /var

jmarrero:
Adds initial validation of /var per: https://github.com/ostreedev/ostree-rs-ext/issues/159#issuecomment-1006033596

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -13,6 +13,7 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 use structopt::StructOpt;
 
+use crate::commit::container_commit;
 use crate::container as ostree_container;
 use crate::container::store::{LayeredImageImporter, PrepareResult};
 use crate::container::{Config, ImageReference, OstreeImageReference, UnencapsulateOptions};
@@ -113,6 +114,9 @@ enum ContainerOpts {
         #[structopt(long)]
         cmd: Option<Vec<String>>,
     },
+
+    #[structopt(alias = "commit")]
+    Commit,
 
     /// Commands for working with (possibly layered, non-encapsulated) container images.
     Image(ContainerImageOpts),
@@ -468,6 +472,7 @@ where
         Opt::Tar(TarOpts::Export(ref opt)) => tar_export(opt),
         Opt::Container(o) => match o {
             ContainerOpts::Info { imgref } => container_info(&imgref).await,
+            ContainerOpts::Commit {} => container_commit().await,
             ContainerOpts::Unencapsulate {
                 repo,
                 imgref,
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -1,0 +1,56 @@
+/// This module contains the functions to implement the commit
+/// procedures as part of building an ostree container image.
+/// https://github.com/ostreedev/ostree-rs-ext/issues/159
+use anyhow::Context;
+use anyhow::Result;
+use std::fs;
+use std::path::Path;
+use tokio::task;
+
+use crate::container_utils::is_ostree_container;
+
+/// Check if there are any files that are not directories and error out if
+/// we find any, /var should not contain any files to commit in a container
+/// as it is where we expect user data to reside.
+fn validate_directories_only(path: &Path, error_count: &mut i32) -> Result<()> {
+    let context = || format!("Validating file: {:?}", path);
+    for entry in fs::read_dir(path).with_context(context)? {
+        let entry = entry?;
+        let path = entry.path();
+
+        let metadata = path.symlink_metadata()?;
+
+        if metadata.is_dir() {
+            validate_directories_only(&path, error_count)?;
+        } else {
+            *error_count += 1;
+            if *error_count < 20 {
+                eprintln!("Found file: {:?}", path)
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Entrypoint to the commit procedures, initially we just
+/// have one validation but we expect more in the future.
+pub(crate) async fn container_commit() -> Result<()> {
+    if is_ostree_container()? {
+        println!("Checking /var for files");
+        let var_path = Path::new("/var");
+
+        let mut error_count = 0;
+
+        task::spawn_blocking(move || -> Result<()> {
+            validate_directories_only(var_path, &mut error_count)
+        })
+        .await??;
+
+        if error_count != 0 {
+            anyhow::bail!("Found content in /var");
+        }
+    } else {
+        anyhow::bail!("Not a container can't commit");
+    }
+    Ok(())
+}
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -35,6 +35,7 @@ pub mod tar;
 pub mod tokio_util;
 
 mod cmdext;
+pub(crate) mod commit;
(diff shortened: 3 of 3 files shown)


lib/src/cli.rs
```diff
@@ -285,6 +292,46 @@ fn tar_export(opts: &ExportOpts) -> Result<()> {
...
+        if is_dirs.is_err() {
+            return Err(anyhow!(
+                "We encountered an error validating '/var', the error is: {:?}",
+                is_dirs.err()
+            ));
+        } else if !is_dirs.unwrap() {
+            return Err(anyhow!("Many files where found under '/var'. No files are allowed under '/var' just directories."));
+        }
+    } else {
+        println!("Not a container can't commit");
```
cgwalters, line 330: I think this should actually be an error.  There is `anyhow::bail!` BTW too which is shorthand for `return Err(anyhow!(...));`

lib/src/cli.rs
```diff
@@ -285,6 +292,46 @@ fn tar_export(opts: &ExportOpts) -> Result<()> {
...
+        let path = entry.path();
+
+        let metadata = fs::metadata(&path)?;
+
+        if metadata.is_dir() {
+            is_only_directories(&path, error_count)?;
+        } else {
+            *error_count += 1;
+            if error_count < &mut 20 {
+                println!("File {:?} found under '/var' found.", path);
```
cgwalters, line 308: Duplicate "found".  Also let's use `eprintln!()` since this is an error.

Maybe: `eprintln!("Found file in /var: {}, path")`
?

cgwalters (commented):
Thanks for working on this!

lib/src/cli.rs
```diff
@@ -285,6 +292,46 @@ fn tar_export(opts: &ExportOpts) -> Result<()> {
...
+            }
+        }
+    }
+    Ok(error_count == &mut 0)
+}
+
+async fn container_commit() -> Result<()> {
```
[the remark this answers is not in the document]
jmarrero: I added the spawn_blocking here: https://github.com/ostreedev/ostree-rs-ext/pull/205/commits/0960531d0c6187b4a59f2bcd638da3ab3d81bc1f  I know you said to bubble up the error so I left the unwrap() do that. However I wonder if doing something like ```if res.is_err() {anyhow::bail!("Found content in /var"{:?})", res.err)}``` is the a better way of just unwrap?

lib/src/cli.rs
```diff
@@ -285,6 +292,46 @@ fn tar_export(opts: &ExportOpts) -> Result<()> {
...
+        let var_path = Path::new("/var/");
+
+        let is_dirs = is_only_directories(var_path, &mut 0);
+        if is_dirs.is_err() {
+            return Err(anyhow!(
+                "We encountered an error validating '/var', the error is: {:?}",
+                is_dirs.err()
+            ));
+        } else if !is_dirs.unwrap() {
+            return Err(anyhow!("Many files where found under '/var'. No files are allowed under '/var' just directories."));
```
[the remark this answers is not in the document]
jmarrero: done

lib/src/cli.rs
jmarrero: the path includes /var. In the latest commit I just changed it to `eprintln!("Found file: {:?}, path")` does that work?

lib/src/cli.rs
jmarrero: done

lib/src/cli.rs
```diff
@@ -285,6 +292,46 @@ fn tar_export(opts: &ExportOpts) -> Result<()> {
...
+        if metadata.is_dir() {
+            is_only_directories(&path, error_count)?;
+        } else {
+            *error_count += 1;
+            if error_count < &mut 20 {
+                println!("File {:?} found under '/var' found.", path);
+            }
+        }
+    }
+    Ok(error_count == &mut 0)
```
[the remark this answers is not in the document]
jmarrero: done

lib/src/cli.rs
```diff
@@ -5,17 +5,21 @@
 //! also exported as a library too, so that projects
 //! such as `rpm-ostree` can directly reuse it.
 
+use anyhow::anyhow;
 use anyhow::Result;
 use ostree::{gio, glib};
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::ffi::OsString;
+use std::fs::{self};
```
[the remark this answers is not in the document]
jmarrero: done

lib/src/commit.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        let mut error_count = 0;
+
+        let res = task::spawn_blocking(move || -> Result<()> {
+            validate_for_directories(var_path, &mut error_count)
+        })
+        .await?;
+        // If get an error during validation we want to fail
+        // and output the error.
+        res.unwrap();
```
cgwalters, line 50: Right, but we should still use `?` for that instead of `unwrap()`.  IOW I think we just need:

```suggestion
        .await??;
```

(This confused me too the first time I encountered it.  It's basically that `spawn_blocking()` can fail, so that's the first `?` - then it calls a function which itself can fail, so that's the second `?`)

lib/src/commit.rs
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Entrypoint to the commit procedures, initially we just
+/// have one validation but we expect more in the future.
+pub(crate) async fn container_commit() -> Result<()> {
+    if is_ostree_container().unwrap() {
```
cgwalters, line 38: Let's also use `?` instead of `unwrap()` here

lib/src/commit.rs
```diff
@@ -0,0 +1,59 @@
...
+        println!("Checking /var for files");
+        let var_path = Path::new("/var/");
```
cgwalters, line 40: I think the trailing `/` here is unnecessary

lib/src/commit.rs
cgwalters: (If we do this we can also drop the `let res = ` I think)

lib/src/commit.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::fs;
+use std::path::Path;
+use tokio::task;
+
+use crate::container_utils::is_ostree_container;
+
+/// Check if there are any files that are not directories and error out if
+/// we find any, /var should not contain any files to commit in a container
+/// as it is where we expect user data to reside.
+fn validate_for_directories(path: &Path, error_count: &mut i32) -> Result<()> {
```
cgwalters, line 15: Maybe `validate_directories_only` or so?

cgwalters (commented):
Looking close to me!  Just have a few smaller comments

lib/src/commit.rs
jmarrero: that worked thanks!

cgwalters (approved):

cgwalters:
I enabled merge for this, but one thing that'd be nice to do after this is extend `ci/integration.sh` to try testing this.

jmarrero:
> I enabled merge for this, but one thing that'd be nice to do after this is extend `ci/integration.sh` to try testing this.

Thanks, I'll add that to my to-do list.

PR #206, opened 2022-01-12
integrationtest: Use mut setters from oci-spec

cgwalters:
Depends https://github.com/containers/oci-spec-rs/pull/87


cgwalters:
OK rebased :surfer: and now pulling from official git.  We'll just need to remember to ask for a new oci-spec-rs release before we do a release, if it hasn't happened by then.

jmarrero (approved):
lgtm

PR #207, opened 2022-01-12
container: Drop `use super::*`

cgwalters:
Came across this in a refactoring.  Glob imports should be used
sparingly.


jmarrero (approved):
lgtm

PR #209, opened 2022-01-12
Use cap-std-ext

cgwalters:
We want to use this code in rpm-ostree too, so let's consider
creating a new shared crate.


cgwalters:
Draft since if we decide to go this route, I need to publish the crate, probably xfer to the coreos/ GH organization etc.

cgwalters:
OK, lifting draft on this; the crate is published!

jmarrero (approved):
lgtm: https://crates.io/crates/cap-std-ext

PR #210, opened 2022-01-19
Add a helper to flatten `Result<Result<T>>` with tokio_util

cgwalters:
I think this is nicer than having `??` which still looks odd to me,
and we can also drop a `map(anyhow::Error::msg)` in once place too.

Prep for further work.

### DIFF
--- a/lib/src/tokio_util.rs
+++ b/lib/src/tokio_util.rs
@@ -1,7 +1,8 @@
 //! Helpers for bridging GLib async/mainloop with Tokio.
 
 use anyhow::Result;
-use futures_util::Future;
+use core::fmt::{Debug, Display};
+use futures_util::{Future, FutureExt};
 use ostree::gio;
 use ostree::prelude::CancellableExt;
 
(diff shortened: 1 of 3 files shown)


lib/src/tokio_util.rs
```diff
@@ -48,6 +49,29 @@ where
...
+
+/// Flatten a nested Result<Result<T>>, defaulting to converting the error type to an `anyhow::Error`.
+/// See https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten
+pub(crate) fn flatten_anyhow<T, E>(r: std::result::Result<Result<T>, E>) -> Result<T>
+where
+    E: Display + Debug + Send + Sync + 'static,
+{
+    match r {
+        Ok(Ok(v)) => Ok(v),
+        Ok(Err(e)) => Err(e),
```
jlebon, line 61: Couldn't this be written:

```suggestion
        Ok(x) => x,
```

?

jlebon (approved):

lib/src/tokio_util.rs
cgwalters: Ah nice!  I started to reply "No because..." then I realized if you just look at it, it's clearly true.

cgwalters: In my defense this code was inspired by https://github.com/rust-lang/rust/issues/70142#issuecomment-750529061 but they're trying to support a "flatten with" which we're hardcoding to be `anyhow!` effectively.  This code is also not as general because it requires the inner to be an `anyhow::Result` today.

jlebon (approved):

Issue #211, opened 2022-01-19
`podman history` segfaults on our generated image


cgwalters:
I thought we'd fixed this with https://github.com/ostreedev/ostree-rs-ext/pull/208 but apparently not.

Stack trace looks like:
```
Goroutine 1 [running]:
panic(0x55a017857c00, 0x55a018610a90)
	/usr/lib/golang/src/runtime/panic.go:1065 +0x565 fp=0xc0004fb778 sp=0xc0004fb6b0 pc=0x55a0160130a5
runtime.panicmem()
	/usr/lib/golang/src/runtime/panic.go:212 +0x5d fp=0xc0004fb798 sp=0xc0004fb778 pc=0x55a01601109d
runtime.sigpanic()
	/usr/lib/golang/src/runtime/signal_unix.go:734 +0x185 fp=0xc0004fb7d0 sp=0xc0004fb798 pc=0x55a01602afc5
github.com/containers/podman/pkg/domain/infra/abi.toDomainHistoryLayer(...)
	/builddir/build/BUILD/podman-3.4.4/_build/src/github.com/containers/podman/pkg/domain/infra/abi/images.go:83
github.com/containers/podman/pkg/domain/infra/abi.(*ImageEngine).History(0xc0000ce198, 0x55a017aec898, 0xc00013a040, 0x7ffde7376545, 0xe, 0x0, 0x0, 0xc0004fbac0)
	/builddir/build/BUILD/podman-3.4.4/_build/src/github.com/containers/podman/pkg/domain/infra/abi/images.go:107 +0x1bf fp=0xc0004fb920 sp=0xc0004fb7d0 pc=0x55a017235d7f
github.com/containers/podman/cmd/podman/images.history(0x55a018635640, 0xc00053ff50, 0x1, 0x1, 0x0, 0x0)
	/builddir/build/BUILD/podman-3.4.4/_build/src/github.com/containers/podman/cmd/podman/images/history.go:81 +0xa9 fp=0xc0004fbd28 sp=0xc0004fb920 pc=0x55a0173917e9
```

Which is clearly a `nil` pointer related to the history.

cgwalters:
We should also patch podman to not segfault in this case of course.

cgwalters:
https://github.com/containers/podman/pull/12936

PR #212, opened 2022-01-19
ocidir: Add creation time into history

cgwalters:
Turns out that `podman history` segfaults if this is missing.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/211


jmarrero (approved):
lgtm

PR #213, opened 2022-01-20
container: Support `/etc/ostree/auth.json` and in `/run` too

cgwalters:
This is related to https://github.com/ostreedev/ostree-rs-ext/issues/121
as well as https://github.com/containers/containers-image-proxy-rs/pull/8
etc.

The CLI code here supports `--authfile`.  However, passing it on
the CLI each time for production use cases pushes complexity to
users.

Add support for global persistent and runtime config files in
`/etc/ostree/auth.json` and `/run/ostree/auth.json`.

Change the default constructor for image pulls to use that by default.

Note that the CLI options override the config defaults.

While we're here, also add `--auth-anonymous` to the CLI, which is
needed to ensure we don't use a config file if present.

### DIFF
--- a/lib/src/globals.rs
+++ b/lib/src/globals.rs
@@ -1,0 +1,62 @@
+//! Global functions.
+
+use super::Result;
+use once_cell::sync::OnceCell;
+use ostree::glib;
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+struct ConfigPaths {
+    persistent: PathBuf,
+    runtime: PathBuf,
+}
+
+/// Get the runtime and persistent config directories.  In the system (root) case, these
+/// system(root) case:  /run/ostree           /etc/ostree
+/// user(nonroot) case: /run/user/$uid/ostree ~/.config/ostree
+fn get_config_paths() -> &'static ConfigPaths {
+    static PATHS: OnceCell<ConfigPaths> = OnceCell::new();
+    PATHS.get_or_init(|| {
+        let mut r = if rustix::process::getuid() == rustix::process::Uid::ROOT {
+            ConfigPaths {
+                persistent: PathBuf::from("/etc"),
+                runtime: PathBuf::from("/run"),
+            }
+        } else {
+            ConfigPaths {
+                persistent: glib::user_config_dir(),
+                runtime: glib::user_runtime_dir(),
+            }
+        };
+        let path = "ostree";
+        r.persistent.push(path);
+        r.runtime.push(path);
+        r
+    })
+}
+
+impl ConfigPaths {
+    /// Return the path and an open fd for a config file, if it exists.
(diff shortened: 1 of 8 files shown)


lib/src/integrationtest.rs
```diff
@@ -61,3 +63,36 @@ pub fn generate_derived_oci(src: impl AsRef<Utf8Path>, dir: impl AsRef<Utf8Path>
...
+    assert_eq!(c.authfile, None);
+    Ok(())
+}
+
+#[cfg(feature = "internal-testing-api")]
+#[context("Running integration tests")]
+pub(crate) fn run_tests() -> Result<()> {
+    // When there's a new integration test to run, add it here.
+    test_proxy_auth()?;
+    println!("integrationtests succeeded.");
```
mkenigs, line 96:
```suggestion
    println!("integration tests succeeded.");
```

lib/src/globals.rs
```diff
@@ -0,0 +1,58 @@
...
+        if let Some(f) = crate::container_utils::open_optional(&persistent)? {
+            return Ok(Some((persistent, f)));
+        }
+        Ok(None)
+    }
+}
+
+/// Return the path to the global container authentication file, if it exists.
+pub(crate) fn get_global_authfile_path() -> Result<Option<PathBuf>> {
+    let paths = get_config_paths();
```
mkenigs, line 54: Not sure how you'd do it in Rust but `paths` feels like it should be C++ `static` since `glib::user_config_dir()` can't change at runtime. And maybe the same thing for `auth.json`, but I'm not sure if you want to support updating that at runtime? If not I was wondering if it could be part of a `ConfigPaths` singleton or something, but I haven't thought that through.

lib/src/globals.rs
```diff
@@ -0,0 +1,58 @@
...
+    };
+    let path = "ostree";
+    r.persistent.push(path);
+    r.runtime.push(path);
+    r
+}
+
+impl ConfigPaths {
+    /// Return the path and an open fd for a config file, if it exists.
+    pub(crate) fn read_file(&self, p: impl AsRef<Path>) -> Result<Option<(PathBuf, File)>> {
```
mkenigs, line 36: Would `open_file` be more accurate?

lib/src/globals.rs
cgwalters: Yep, makes sense, I switched to a [OnceCell](https://docs.rs/once_cell/1.9.0/once_cell/#lazy-initialized-global-data).

cgwalters:
Thanks for reviewing!

lib/src/globals.rs
mkenigs: *reads docs* ... :exploding_head:

mkenigs (approved):

PR #214, opened 2022-01-24
Use new correct `impl Default` for oci bits

cgwalters:
Now that https://github.com/containers/oci-spec-rs/pull/90/commits/6e65562e3d70b66656598e6a567dc71e80dd0f56
landed, we can just depend on it and not have a hardcoded
architecture mapping here.


jmarrero (approved):
lgtm

PR #215, opened 2022-01-24
Add and use an API to require being in an ostree container

cgwalters:
utils: Add an API to require being in an ostree container

Most of our callers actually want this.

---

lib/commit: Use `require_ostree_container()` API

---


jmarrero (approved):
lgtm

PR #216, opened 2022-01-24
Switch to new `require_rev` API

cgwalters:
I added this back in
https://github.com/ostreedev/ostree-rs/pull/35/commits/f8aa658d17af9c13702a20ea8d4030cc34121bc8
and only recently remembered about it when modifying some other
code.

We need some sort of automatic reminder system for "remember
to use this new API from your dependency".


jmarrero (approved):
lgtm

PR #217, opened 2022-01-25
tar/export: add additional mandatory ostree repo directories

lucab:
This enhances the logic for tar-exporting, so that
libostree can more easily operate on the repo.
It also contains a couple of bugfixes, and an overall unit-test
for the repo skeleton.

/cc @cgwalters

### DIFF
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -111,7 +111,6 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
         if self.wrote_initdirs {
             return Ok(());
         }
-        self.wrote_initdirs = true;
 
         let objdir: Utf8PathBuf = format!("{}/repo/objects", OSTREEDIR).into();
         // Add all parent directories
@@ -132,6 +131,16 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
             let path: Utf8PathBuf = format!("{}/{:02x}", objdir, d).into();
             self.append_default_dir(&path)?;
         }
+        // Tmp subdirectories
+        for d in ["tmp", "tmp/cache"] {
+            let path: Utf8PathBuf = format!("{}/repo/{}", OSTREEDIR, d).into();
+            self.append_default_dir(&path)?;
+        }
+        // Refs subdirectories
+        for d in ["refs", "refs/heads", "refs/mirrors", "refs/remotes"] {
+            let path: Utf8PathBuf = format!("{}/repo/{}", OSTREEDIR, d).into();
+            self.append_default_dir(&path)?;
+        }
 
         // The special `repo/xattrs` directory used only in our tar serialization.
         let path: Utf8PathBuf = format!("{}/repo/xattrs", OSTREEDIR).into();
@@ -150,6 +159,7 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
         self.out
             .append_data(&mut h, path, std::io::Cursor::new(REPO_CONFIG))?;
 
+        self.wrote_initdirs = true;
         Ok(())
     }
 
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -127,7 +127,7 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
             self.append_default_dir(path)?;
         }
         // Object subdirectories
-        for d in 0..0xFF {
+        for d in 0..=0xFF {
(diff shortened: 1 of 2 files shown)


lib/src/tar/export.rs
```diff
@@ -455,4 +466,47 @@ mod tests {
...
+        let mut archive = {
+            let mut builder = tar::Builder::new(vec![]);
+            let mut writer = OstreeTarWriter::new(&mut builder, ExportOptions::default());
+            writer.write_repo_structure().unwrap();
+            assert!(writer.wrote_initdirs);
+            let buf = builder.into_inner().unwrap();
+            tar::Archive::new(std::io::Cursor::new(buf))
+        };
+
+        for entry in archive.entries().unwrap() {
```
cgwalters, line 503: There's a similar test here https://github.com/ostreedev/ostree-rs-ext/blob/392d5ed2d46c0fe7f966efc7705730356e620fa4/lib/tests/it/main.rs#L286 - may make sense to unify?

lib/src/tar/export.rs
```diff
@@ -125,7 +125,7 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
             self.append_default_dir(path)?;
         }
         // Object subdirectories
-        for d in 0..0xFF {
+        for d in 0..=0xFF {
```
cgwalters, line 128: Eeek!  Oops!

lib/src/tar/export.rs
```diff
@@ -403,18 +407,24 @@ fn impl_export<W: std::io::Write>(
...
-#[derive(Debug, Default, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct ExportOptions {
     /// Format version; must be 0 or 1.
     pub format_version: u32,
 }
 
+impl Default for ExportOptions {
+    fn default() -> Self {
+        Self { format_version: 1 }
```
cgwalters, line 424: I am thinking perhaps we should wait a bit before pulling the trigger on this.  The new export format can only be read by rpm-ostree v2022.1, not v2021.14 (I think).

Or well...hmm, OTOH maybe this is a good time to make the change because we'll be doing a new semver bump too.

Hopefully people using this on the client side can quickly update. So...OK!

lib/src/tar/export.rs
cgwalters: Yep no worries.  Often now I'm starting to write integration tests by default - using unit tests only for testing internal APIs or more trivial things.

A big benefit of integration tests is that one is also testing whether all the APIs are public correctly, etc.

cgwalters: (Perhaps it's also time to break up `main.rs` into separate test modules)

cgwalters (approved):

lib/tests/it/main.rs
```diff
@@ -310,8 +326,26 @@ fn test_tar_export_structure() -> Result<()> {
...
+        ("sysroot/ostree/repo/refs", Directory, 0o755),
+        ("sysroot/ostree/repo/refs/heads", Directory, 0o755),
+        ("sysroot/ostree/repo/refs/mirrors", Directory, 0o755),
+        ("sysroot/ostree/repo/refs/remotes", Directory, 0o755),
+        ("sysroot/ostree/repo/tmp", Directory, 0o755),
+        ("sysroot/ostree/repo/tmp/cache", Directory, 0o755),
+        (
+            "sysroot/ostree/repo/xattrs",
+            Directory,
+            libc::S_IFDIR | 0o755,
```
lucab, line 346: This test fails, so I think we have a split-brain when we generate the tar content:
 * some directories like `sysroot/ostree/repo/xattrs` have mode `0o755`
 * some directories like `usr` have mode `libc::S_IFDIR | 0o755`

My current reading of `tar` specs is that the correct form is the first one, maybe?

cgwalters: Right.  Today in ostree we end up mixing format and permissions in the mode bits.  I think this was a mistake.

Agree that our tar serialization should not inject mode bits into the permissions members, because tar already has a distinct field for that.

cgwalters: Do you want to make this fix, or me?

cgwalters: OK I pushed a commit to ignore it in tests

cgwalters (approved):

lib/tests/it/main.rs
lucab: Thanks for that. It was late evening already when I found out that the import path was also relying on modebits mixup.
I finished squashing the (hopefully) last buggy codepath today, PR at https://github.com/ostreedev/ostree-rs-ext/pull/220.

PR #219, opened 2022-01-25
Release 0.6.1

cgwalters:
build: Use released oci-spec

See https://github.com/containers/oci-spec-rs/pull/94

---

lib: Bump rustix and cap-std-ext

cap-std-ext needs the latest rustix.

---

Release 0.6.1

I believe this is all semver compat; we have a big change queued
up for the layer split work which should come after this.

---


lucab:
If possible, it would be good to sneak https://github.com/ostreedev/ostree-rs-ext/pull/220 in before the release, so that exported tar artifacts are a bit less broken.

cgwalters:
Yeah, it's also time to transfer cap-std-ext to coreos/ and have it require peer review I think.

cgwalters:
OK, respun with https://crates.io/crates/cap-std-ext/0.23.0

lucab (approved):

lucab:
@cgwalters ugh sorry, feel to just revert it for now and release without that. I'll land better logic after that release.

PR #220, opened 2022-01-26
tar/export: fix mode bits in tar archive

lucab:
This fixes the tar exporting logic to avoid leaking filetype bits
into mode field. The tar header already has a distinct field to
identify entry type, thus mode bits should only convey permissions.

Ref: https://github.com/ostreedev/ostree-rs-ext/pull/217#discussion_r791942496

### DIFF
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -281,7 +281,7 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
         h.set_uid(meta.attribute_uint32("unix::uid") as u64);
         h.set_gid(meta.attribute_uint32("unix::gid") as u64);
         let mode = meta.attribute_uint32("unix::mode");
-        h.set_mode(mode);
+        h.set_mode(mode & !libc::S_IFMT);
         let mut target_header = h.clone();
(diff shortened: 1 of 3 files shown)


lucab:
@cgwalters I think you have a semver-breaking change in `cap-std-ext/0.1.2`. You should probably yank, after that re-triggering CI should (in theory) pass here.

cgwalters:
Sorry, done!

lib/src/tar/export.rs
```diff
@@ -281,7 +281,7 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
         h.set_uid(meta.attribute_uint32("unix::uid") as u64);
         h.set_gid(meta.attribute_uint32("unix::gid") as u64);
         let mode = meta.attribute_uint32("unix::mode");
-        h.set_mode(mode);
+        h.set_mode(mode & 0o7777);
```
cgwalters, line 284: Let's use `& !libc::S_IFMT` for greppability?

lucab: Ack, done!

cgwalters (approved):

PR #222, opened 2022-01-26
tar: revert to previous mode bits, fix importing

lucab:
This reverts https://github.com/ostreedev/ostree-rs-ext/pull/220.
But it keeps a patch which should be fixing the underlying import bug.


cgwalters:
Let's try https://github.com/ostreedev/ostree-rs-ext/pull/223 instead?  I think since we do have the format version it'd be good to try to fix this in the new one.

PR #223, opened 2022-01-26
tar/export: Continue injecting mode bits in format version 0

cgwalters:
This ensures compatibility with older clients, which we need
to pass the rpm-ostree upgrade test.


cgwalters:
This passes unit tests, going to test that it helps pass the rpm-ostree upgrade test now.

cgwalters:
OK yep, this passes the upgrade test :tada:

lucab (approved):

PR #228, opened 2022-02-01
lib/tar: Simplify import code, minor comment tweaks

cgwalters:
I was just reading this and noticed we can simplify it.


jmarrero (approved):
lgtm

PR #229, opened 2022-02-01
Bump cap-std to 0.24

cgwalters:
This picks up various fixes, but most critically a new rustix
with https://github.com/bytecodealliance/rustix/pull/186


jmarrero (approved):
lgtm

PR #230, opened 2022-02-02
ocidir: Use `chrono`, not `glib` for time formatting

cgwalters:
Sadly RHEL8's glib is too old for this.  Switch to using chrono.


jmarrero (approved):
lgtm

PR #233, opened 2022-02-03
ocidir: More direct use of direct `_mut()` methods

cgwalters:
Followup to https://github.com/ostreedev/ostree-rs-ext/pull/206/commits/e0c4c849d9901e0de03744a6200117424f1b7158

This is notably more efficient than deep cloning all the elements.

Just noticed while reading the code for an unrelated reason.


jmarrero (approved):
lgtm

PR #234, opened 2022-02-03
container/encapsulate: Support copying commit metadata keys

cgwalters:
For https://github.com/coreos/coreos-assembler/issues/2685
we want to copy e.g. `rpmostree.input-hash` into the container image.
Extend the `ExportOpts` struct to support this, and also expose
it via the CLI, e.g.
`ostree container encapsulate --copymeta=rpmostree.input-hash ...`.

And while I was thinking about this...we should by default copy
some core ostree keys, such as `ostree.bootable` and `ostree.linux`
since they are key pieces of metadata.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -110,6 +110,10 @@ enum ContainerOpts {
         #[structopt(name = "label", long, short)]
         labels: Vec<String>,
 
+        /// Propagate an OSTree commit metadata key to container label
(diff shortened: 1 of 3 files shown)


cgwalters:
OK yeah, there's some bug in `cargo fmt` where it's non-idempotent in this case of content inside the `indoc!` macro.

jlebon (approved):

PR #236, opened 2022-02-04
Hard depend on `skopeo copy --digestfile`

cgwalters:
This code predated our dependency on `experimental-image-proxy`,
so we can now safely carry a hard dependency on a new enough
skopeo and drop the pile of hacks here.


jmarrero (approved):
lgtm

PR #237, opened 2022-02-08
[RFC] tar/v1: introduce an object type for split xattrs

lucab:
This reworks the v1 tar format (and the 'bare-split-xattrs' mode too),
moving xattrs content under the usual sharded object store.
In particular, this adds two new object types.

`.file-xattrs` are regular files storing xattrs content, encoded as
GVariant. Each object is keyed by the checksum of its content, allowing
for multiple references.

`.file-xattrs-link` are symlinks which are associated to `.file` objects.
Each object is keyed by the same checksum of the corresponding file
object. The target of the symlink is an existing `file-xattrs` object.


lib/src/tar/import.rs
```diff
@@ -139,6 +146,46 @@ fn parse_checksum(parent: &str, name: &Utf8Path) -> Result<String> {
...
+        bail!("Invalid objpath {} for {}", objtype, path);
+    }
+
+    // Reassemble the target checksum and validate it.
+    let target = {
+        let basename = fname.as_str().trim_end_matches(".file-xattrs");
+        let checksum = format!("{}{}", parentname, basename);
+        validate_sha256(&checksum)?;
+        checksum
+    };
```
cgwalters, line 184: I don't quite get the value of the nested scope here.  Hmm, would it be cleaner if e.g. `validate_sha256()` took ownership and returned it?  Then it could be:
`Ok(validate_sha256(format!("{}{}", parentname, basename)))`
or so.

cgwalters (approved):
Overall LGTM!  Some minor nonblocking nits.

lib/src/tar/export.rs
```diff
@@ -250,43 +279,87 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
...
         let xattrs_data = xattrs_data.as_ref();
         if xattrs_data.is_empty() {
-            return Ok(None);
+            return Ok(false);
         }
 
         let mut h = tar::Header::new_gnu();
+        h.set_uid(0);
+        h.set_gid(0);
         h.set_mode(0o644);
```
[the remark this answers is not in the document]
lucab: Ack, I introduced two helper for adding regular files and hardlinks in the ostree repository.

lib/src/tar/import.rs
lucab: It just mirrors my mental logic organization, as the matching comment before that suggests. I removed this nesting to avoid doubts.

I also updated the validation function to transfer ownership as suggested.

lucab:
Found a couple of issues with this PR while cross-testing the `libostree` changes. Followups:
 * https://github.com/ostreedev/ostree-rs-ext/pull/240
 * https://github.com/ostreedev/ostree-rs-ext/pull/241

cgwalters:
I think this one broke imports from the current v0 format:
```
$ error: Importing ostree-unverified-image:docker://quay.io/coreos-assembler/fcos:testing-devel: Parsing blob sha256:16fae988564d496e5caf15aefbe41afaa722240e6b130ba535a6515ba3bdadad: Importing object 0e/aa1ed36ca97a563a80f027f0de643c5db36d6a98dd857b609ae429aca7aafc.file.xattrs: Processing xattrs reference: Found previous dangling xattrs for file object '936f751855ebe4210d3fb59d3ffca58c2270dc4d6d89c4507abe40eb530ddd69'
$
```

cgwalters:
Well, that problem was actually https://github.com/ostreedev/ostree-rs-ext/pull/242

PR #239, opened 2022-02-10
cli: Drop use of `select`

cgwalters:
I saw https://blog.yoshuawuyts.com/futures-concurrency-3/
go by and decided to audit our use of `select!` - I am definitely
not an expert but my understanding is the CLI use case is buggy,
but the other one is not.

Rewrite the CLI case to operate on an explicit stream of values.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -6,16 +6,19 @@
 //! such as `rpm-ostree` can directly reuse it.
 
 use anyhow::Result;
+use futures_util::FutureExt;
 use ostree::{gio, glib};
+use std::borrow::Borrow;
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use structopt::StructOpt;
+use tokio_stream::StreamExt;
 
 use crate::commit::container_commit;
-use crate::container as ostree_container;
 use crate::container::store::{LayeredImageImporter, PrepareResult};
+use crate::container::{self as ostree_container, UnencapsulationProgress};
 use crate::container::{Config, ImageReference, OstreeImageReference, UnencapsulateOptions};
 
 fn parse_imgref(s: &str) -> Result<OstreeImageReference> {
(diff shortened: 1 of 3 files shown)


lib/src/cli.rs
```diff
@@ -324,26 +332,32 @@ async fn container_import(
...
                 if let Some(pb) = pb.as_ref() {
                     pb.set_message(format!("Processed: {}", indicatif::HumanBytes(n)));
                 }
             }
-            import = &mut import => {
+            ProgressOrFinish::Finished(import) => {
                 if let Some(pb) = pb.as_ref() {
                     pb.finish();
                 }
-                break import?;
```
lucab, line 342: I think that dropping this `break` may now result in `pb.set_message()` possibly being called after a `pb.finish()`, depending on the ordering of items in the merged stream. I'm not too familiar with `indicatif`, but that behavior sounds risky.

If this was an intended change with the aim of fully driving the stream to completion, then I think the `Progress => {...}` arm should also be augmented with a `if !pb.is_finished()` check.

cgwalters: Great catch!  I chose to move the progress bar completion outside of the stream processing which seems simplest.

Though, it seems like this pattern where one element of the stream is a singleton could re-occur, and we could find/create a helper for this.  But...eh.

I'm still not 100% sure that the original use of `select!` is buggy...we previously did this:

````
let import = crate::container::unencapsulate(repo, imgref, Some(opts));
tokio::pin!(import);
```

which seems to me should prevent the value from being dropped and re-created.  

(A bit of time passes)

OK yes, [this linked blog entry](https://tomaka.medium.com/a-look-back-at-asynchronous-rust-d54d63934a1c#3008) has:

> Rewrite the select! to not destroy the future. [Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d0c95a33aab6859bfda3405e9b29b5ef). This is arguably the best solution in that specific situation, but it can sometimes introduce a lot of complexity, for example if you want to re-create the future with a different File when the socket receives a message.

And indeed, the linked fix there is to pin the future, which is exactly what we were doing.

So...dunno.  I guess I'd probably still vote for merging this.  The code is perhaps a bit clearer at least, and `select!` is still a trap - but ISTM that a linter could warn about any "nontrivial" non-pinned futures somehow or so?

lucab: I think both versions are correct, but as the blogpost mentions this kind of code is hard to reason about and review.
I'm happy to land this new macro-less version.

PR #240, opened 2022-02-11
tar/export: export empty xattrs in v1 format

lucab:
This makes sure that xattrs content is always exported in v1 format,
even the empty attributes set. Previous behaviour is retained for v0,
for compatibility.


cgwalters (approved):
OK as is but we'll probably need to add a test case for the no-xattrs case in the future.

PR #242, opened 2022-02-11
tar/import: Don't crash on extant object

cgwalters:
This bug may have existed before, but basically we need
to handle the case where an object already exists in the repo;
we were erroring out in this case because we retained the queued
xattrs entry.

### DIFF
--- a/lib/src/tar/import.rs
+++ b/lib/src/tar/import.rs
@@ -324,12 +324,6 @@ impl Importer {
         checksum: &str,
         cancellable: Option<&gio::Cancellable>,
     ) -> Result<()> {
-        if self
-            .repo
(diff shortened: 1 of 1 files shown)


cgwalters:
I did start to write a test for this but ran into another pile of technical debt around our test suite that I will try to tackle separately.

lucab (approved):

lucab:
LGTM, although I'm surprised the input tar has multiple file objects with the same checksum.

cgwalters:
> LGTM, although I'm surprised the input tar has multiple file objects with the same checksum.

No, the scenario here is that I was pulling into a repo that already had a FCOS build so there were some objects already present.

PR #244, opened 2022-02-14
testsuite: Port to `sh-inline` 0.2

cgwalters:
Prep for further work.

I just cut a new release of `sh-inline` - we now have a `cap-std-ext`
feature which lets us use `bash_in!` which takes a reference
to a directory fd.

Also of note, passed variables are now shell-like `${foo}` instead
of substituted.

This is significantly less fragile than having our script do a
`cd` at the top.

This also starts us down the path to porting the framework to use
`cap-std`, e.g. the fixture uses the new `open_at_dir()` API from
https://github.com/ostreedev/ostree-rs/commit/e1fa1da4cafad874a6e58fbea326ded5255e0cdd

### DIFF
--- a/lib/tests/it/fixture.rs
+++ b/lib/tests/it/fixture.rs
@@ -1,10 +1,15 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
+use cap_std::fs::Dir;
+use cap_std_ext::prelude::CapStdExtCommandExt;
 use fn_error_context::context;
 use indoc::indoc;
+use ostree::cap_std;
 use ostree_ext::gio;
-use sh_inline::bash;
+use sh_inline::bash_in;
 use std::convert::TryInto;
+use std::process::Stdio;
+use std::sync::Arc;
 
 const OSTREE_GPG_HOME: &[u8] = include_bytes!("fixtures/ostree-gpg-test-home.tar.gz");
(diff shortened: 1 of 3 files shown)


lib/tests/it/fixture.rs
```diff
@@ -58,18 +69,18 @@ impl Fixture {
...
         let srcpath = &repotmp.join("exampleos-v1.tar.zst");
         std::fs::write(srcpath, EXAMPLEOS_V1)?;
         let srcpath = srcpath.as_str();
-        let repopath = repopath.as_str();
         let testref = TESTREF;
-        bash!(
-            "ostree --repo={repopath} commit -b {testref} --no-bindings --tree=tar={srcpath}",
+        bash_in!(
+            self.dir.open_dir("src")?,
+            "ostree --repo=repo commit -b ${testref} --no-bindings --tree=tar=${srcpath}",
```
jmarrero, line 82: I am confused on where is repo declared here. I see the tests with `--repo=repo` too but Based on te line you removed I would expect some variable here too?

cgwalters: Good question!  The change here is because we use `bash_in!` which changes the working directory for the bash code, we can just use relative paths there now.

jmarrero: Ahh makes sense.

jmarrero (approved):
lgtm

PR #245, opened 2022-02-14
A few more test suite patches

cgwalters:
tests: Split off fixture base prep from repo initialization

---

tests: Port to use more `bash_in!` and drop absolute paths

---

tests: Drop src absolute path

This patch series is staring to feel like spraying Windex on a dirty
window, and watching it get cleaner.

---

tests: Use cap-std in more places

---

tests: Move initial commit into fixture

---


jmarrero (approved):
lgtm

PR #247, opened 2022-02-15
cli: Use structopt's `TryFrom` support for parsing `ostree::Repo`

cgwalters:
It's stuff like this that is just *so elegant* in Rust.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -7,7 +7,7 @@
 
 use anyhow::Result;
 use futures_util::FutureExt;
-use ostree::{gio, glib};
+use ostree::{cap_std, gio, glib};
 use std::borrow::Borrow;
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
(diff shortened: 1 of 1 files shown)


lib/src/cli.rs
```diff
@@ -6,6 +6,7 @@
 //! such as `rpm-ostree` can directly reuse it.
 
 use anyhow::Result;
+use cap_std_ext::cap_std;
```
lucab, line 9: I think you'd want `ostree::cap_std` here, in order to match what `open_at_dir` wants.

At the time of new releases, `cap_std_ext::cap_std` and `ostree::cap_std` may be at incompatible versions.

cgwalters: You're definitely right in this instance, but in general we use both `cap-std-ext` and `ostree`, and I don't think we really want to have two different versions of cap-std in the dependency graph.

IOW semver bumps there are going to involve churn for us no matter what I think.

lucab (approved):

PR #250, opened 2022-02-16
tests: Add much beefed up alternative path for test framework

cgwalters:
Depends: https://github.com/ostreedev/ostree-rs-ext/pull/248

---

tests: Add much beefed up alternative path for test framework

The main thing I want to get away from is hardcoded tarball fixtures.

I do want the commit digests we write to be reproducible, and
I also want the unit tests to work on a non-SELinux system.

This patch adds an alternative flow to the text fixture via `new_v1()`
that uses the native ostree APIs to write and generate commits.  We
mock up the SELinux labeling (and support labeling disabled).

I plan to drop the old path and hardcoded tar fixtures once this lands.

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -42,23 +42,24 @@ async fn test_tar_import_empty() -> Result<()> {
 
 #[tokio::test]
 async fn test_tar_export_reproducible() -> Result<()> {
-    let fixture = Fixture::new()?;
-    let (_, rev) = fixture
-        .srcrepo
-        .read_commit(fixture.testref(), gio::NONE_CANCELLABLE)?;
-    let export1 = {
-        let mut h = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
-        ostree_ext::tar::export_commit(&fixture.srcrepo, rev.as_str(), &mut h, None)?;
-        h.finish()?
-    };
-    // Artificial delay to flush out mtimes (one second granularity baseline, plus another 100ms for good measure).
-    std::thread::sleep(std::time::Duration::from_millis(1100));
-    let export2 = {
-        let mut h = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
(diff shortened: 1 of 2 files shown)


lib/tests/it/main.rs
```diff
@@ -67,30 +42,40 @@ async fn test_tar_import_empty() -> Result<()> {
...
+            h.finish()?
+        };
+        // Artificial delay to flush out mtimes (one second granularity baseline, plus another 100ms for good measure).
+        std::thread::sleep(std::time::Duration::from_millis(1100));
+        let export2 = {
+            let mut h = openssl::hash::Hasher::new(openssl::hash::MessageDigest::sha256())?;
+            ostree_ext::tar::export_commit(&fixture.srcrepo, rev.as_str(), &mut h, None)?;
+            h.finish()?
+        };
+        assert_eq!(*export1, *export2);
```
lucab, line 61: Two alternatives options, just as suggestions for your consideration:
 1. Hardcode a list of digests from the current baseline, to ensure the exports do not unexpectedly diverge over time
 2. Same as 1, but also also committing the fixtures (which are programmatically generated by the testsuite) for easier debugging on regressions

lucab (commented):
Overall I really like the direction of having fixtures programmatically generated.
However I have a concern that we may be losing checks that the import/export  logic don't diverge in a breaking way over time. See inline comment.

lib/tests/it/main.rs
cgwalters: One problem with hardcoding the digests of the tar stream is that it can change if the bitstream from [tar-rs](https://github.com/alexcrichton/tar-rs) changes.

Now, on one hand we need to be aware of this, because bitstream generation changes there will actually be very noticeable after we implement https://github.com/ostreedev/ostree-rs-ext/pull/123 because it will force clients to redownload *all* layers.

cgwalters:
OK, this is isn't done, but let's merge it as a checkpoint - the new framework is opt-in behind `new_v1()` and only used in one test.  Porting all the tests I think should be done as a separate pass.

lucab (approved):

PR #251, opened 2022-02-18
lib/src/tar/write: make sure we add the links when filtering the tar

jmarrero:


cgwalters (commented):
Awesome, thanks so much for tracking this down!

cgwalters (approved):
Nice!

PR #252, opened 2022-02-22
ostree-ext: release 0.6.4

lucab:
Changes:
 * tar/export: add 'state' and 'extensions' repo subdirs
 * tests: Add much beefed up alternative path for test framework
 * lib/src/tar/write: make sure we add the links when filtering the tar
 * tests: two more small test suite patches
 * tar: Add a debug function to filter tar
 * cli: Use structopt's `TryFrom` support for parsing `ostree::Repo`
 * tests: A few more test suite patches
 * testsuite: Port to `sh-inline` 0.2
 * tar/import: Don't crash on extant object
 * tests: Factor out test fixture as a separate module
 * tar/export: use full path as link target
 * tar/export: export empty xattrs in v1 format
 * cli: Drop use of `select`
 * tar/v1: introduce object types for split xattrs
 * container: Hard depend on `skopeo copy --digestfile`
 * container/encapsulate: Support copying commit metadata keys
 * ocidir: More direct use of direct `_mut()` methods

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.6.3"
+version = "0.6.4"
 
 [dependencies]
 anyhow = "1.0"


lib/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.6.3"
+version = "0.7.0"
```
cgwalters, line 9: Did we do any semver-incompatible changes?  I didn't think so...

Not opposed to a bump, but I think on general principle it's better to only do it when needed.

lucab: The full diff is at <https://github.com/ostreedev/ostree-rs-ext/compare/ostree-ext-v0.6.3...main>.

It is my understanding that we had breaking change to the public `ExportOpts` type at <https://github.com/ostreedev/ostree-rs-ext/pull/234/files#diff-1008fb4c697b3e2fbd5ed48c12cd3dfde815f20d76817f3f584f6421a5515e4eR232>. I based my versioning decision based on this, but let me know in case I misjudged it.

lucab:
This needs an update to the crates.io permissions in order to proceed:
```
cargo owner --add github:ostreedev:crates ostree-ext
```

lib/Cargo.toml
cgwalters: The API consumers should be constructing [using Default::default() to handle unknown fields](https://github.com/ostreedev/ostree-rs-ext/blob/aeee94e53a9f66272bb4e66b5da4ee6c62d9125c/lib/src/cli.rs#L411).  

Now again to be clear, I don't care a lot about this *particular* time, but I think as a general rule we should be trying to handle this well, because it will matter more once the features in this crate get towards 1.0.

Relatedly, I think we should be tracking semver-incompatible changes via e.g. a `[breaking]` header.

Basically though for now if rpm-ostree doesn't require any changes to build with this release, then let's not bump semver.

cgwalters:
> This needs an update to the crates.io permissions in order to proceed:

```
$ cargo owner --add github:ostreedev:crates ostree-ext

    Updating crates.io index
       Owner team github:ostreedev:crates has been added as an owner of crate ostree-ext
$
```

lib/Cargo.toml
cgwalters: I created a [semver-break label](https://github.com/ostreedev/ostree-rs-ext/labels/semver-break) and the big layer rework unifies the import APIs, which is definitely a semver break.

lucab: Following your decision I did bring this back to `0.6.4` according to the definition of "rpm-ostree will be fine with it".

For clarity, to my understanding the one above is indeed a semver-breaking change, no matter what consumers do.

On the 1.0 topic, this library re-exports several other crates (gio/glib/ostree/oci/etc.) and any incompatible bump of those will reflect to this library too (under standard semver rules).

cgwalters: > For clarity, to my understanding the one above is indeed a semver-breaking change, no matter what consumers do.

Hmm.  Do you have a reference for that?  What would be the alternative?  The builder pattern?

cgwalters (approved):

lib/Cargo.toml
lucab: I tried to confirm this via [semverver](https://github.com/rust-lang/rust-semverver) this morning, but unfortunately due to its old-nightly requirement it can't build our crate (maybe it could with enough hammering, but I stopped after a few failed attempts).

The most explicit reference I found about this is <https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-adding-a-public-field-when-no-private-field-exists>.

If you are looking for a general alternative pattern, I think the above is a hint towards adding one phantom private field, but I'm not fully sure about that.

(But the larger elephant in the room is that there are many change points outside of our control too, so I'm not sure how realistic it is to pursue strong stability guarantees at this point. With that in mind, I'd approach the whole bumping topic more liberally)

PR #253, opened 2022-02-23
tar/export: add 'state' and 'extensions' repo subdirs

lucab:
This adds two missing standard repo subdirectories, `state` and `extensions`. The former is required for proper `fsck` usage.

It also fixes an unrelated test which is currently not building, due to a merge overlap.


jmarrero (approved):
lgtm

PR #254, opened 2022-02-24
tar/export: Fix duplicate xattrs in tar stream

cgwalters:
The containers/storage stack fails if it sees duplicate files.

Fix the bug introduced in the format version rework that passed
the object checksum instead of the xattrs checksum into the "seen"
map.

While we have the patient open, I noticed that we can optimize
things a bit and pass ownership of the checksum into the map.

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -10,7 +10,7 @@ use ostree_ext::container::{
 use ostree_ext::tar::TarImportOptions;
 use ostree_ext::{gio, glib};
 use sh_inline::bash_in;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::process::Command;
 
 use fixture::Fixture;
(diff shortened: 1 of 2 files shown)


cgwalters:
https://github.com/coreos/rpm-ostree/pull/3469#issuecomment-1050135011

lib/tests/it/main.rs
```diff
@@ -156,11 +156,16 @@ fn validate_tar_expected<T: std::io::Read>(
...
         expected.into_iter().map(|exp| (exp.path, exp)).collect();
     let entries = t.map(|e| e.unwrap());
+    let mut seen_paths = HashSet::new();
     // Verify we're injecting directories, fixes the absence of `/tmp` in our
     // images for example.
     for entry in entries {
         let header = entry.header();
         let entry_path = entry.path().unwrap().to_string_lossy().into_owned();
+        if seen_paths.contains(&entry_path) {
+            anyhow::bail!("Duplicate path: {}", entry_path);
```
cgwalters, line 166: I did also cross-verify this test fails without the change to `export.rs`.

jmarrero (approved):
lgtm

PR #256, opened 2022-02-24
Fix a few clippy lints from new Rust 1.59 version

cgwalters:
Fix a few clippy lints from new Rust 1.59 version

---

### DIFF
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -118,7 +118,7 @@ fn build_oci(
 
     labels.insert(OSTREE_COMMIT_LABEL.into(), commit.into());
 
-    for (k, v) in config.labels.iter().map(|k| k.iter()).flatten() {
+    for (k, v) in config.labels.iter().flat_map(|k| k.iter()) {
         labels.insert(k.into(), v.into());
     }
     // Lookup the cmd embedded in commit metadata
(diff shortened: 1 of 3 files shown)


lib/src/container/encapsulate.rs
```diff
@@ -118,13 +118,13 @@ fn build_oci(
...
 
-    for (k, v) in config.labels.iter().map(|k| k.iter()).flatten() {
+    for (k, v) in config.labels.iter().flat_map(|k| k.iter()) {
         labels.insert(k.into(), v.into());
     }
     // Lookup the cmd embedded in commit metadata
     let cmd = commit_meta.lookup::<Vec<String>>(ostree::COMMIT_META_CONTAINER_CMD)?;
     // But support it being overridden by CLI options
-    let cmd = config.cmd.as_ref().or_else(|| cmd.as_ref());
+    let cmd = config.cmd.as_ref().or(cmd.as_ref());
```
jlebon, line 127: Clippy is not liking this change.

cgwalters: Ah awesome, old clippy doesn't like new code, new clippy doesn't like old code.  Well, we can deal with this one when we update the CI pinned clippy.

cgwalters: Took this to https://github.com/rust-lang/rust-clippy/pull/7639#issuecomment-1050340564

jlebon (approved):

PR #257, opened 2022-02-28
ci: Install ostree from coreos/continuous copr

cgwalters:
ci: Drop hack to build skopeo

A new enough version is now in Fedora.

---

ci: Install ostree from coreos/continuous copr

Mainly to pick up https://github.com/ostreedev/ostree/pull/2548

What we really want of course is to have FCOS and build containers
that consume the continuous directly.  This is just dipping our toes
in that water.

---

### DIFF
--- a/ci/installdeps.sh
+++ b/ci/installdeps.sh
@@ -6,3 +6,20 @@ dnf -y --enablerepo=updates-testing update ostree-devel
 
 # Our tests depend on this
 dnf -y install skopeo
+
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
> CI can't find skopeo

Ah, right it's not in the buildroot, we were actually *always* building it.  Changed to just install current Fedora version.

jmarrero (approved):
lgtm

PR #258, opened 2022-02-28
tests: More fixture rework, port test-diff to new path

cgwalters:
Currently the fixture is just built during `cargo test`; to help
debug it I wanted to support it being part of our main shared
library in the "integrationtest" path.  For example I want to add
a CLI method to stand up the fixture directory (but not delete it).

But the other big thing going on here is that we now support updating
commits fully in memory, and so the diff API is ported to use this.

This requires https://github.com/ostreedev/ostree/pull/2548


cgwalters:
After this the next step is to port everything to the new fixture, and drop the hardcoded tarballs.

jmarrero (approved):
lgtm

PR #259, opened 2022-03-07
ci/installdeps.sh: Clean up to enable repos consistently

cgwalters:
We should have a clear two-step flow where we enable repos (updates-testing
and the CoreOS/continuous COPR) and then install packages.

This way we also get skopeo from updates-testing which a future
PR will need.


jmarrero (approved):
lgtm

PR #260, opened 2022-03-07
Add new `objectsource` module

cgwalters:
In ostree we aim to provide generic mechanisms that can be
consumed by any package or build system.

Hence we often use the term "component" instead of "package".

This new `objectsource` module is an abstraction over basic metadata
for a component/package, currently name, identifier, and last change time.

This will be used for splitting up a single ostree commit back
into "chunks" or container image layers, grouping objects
that come from the same component together.

https://github.com/ostreedev/ostree-rs-ext/issues/69

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -37,7 +37,9 @@ pub mod tar;
 pub mod tokio_util;
 
 pub(crate) mod commit;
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

PR #261, opened 2022-03-07
bootabletree: New module with helpers for bootable ostree commits

cgwalters:
Right now this only offers an API to find the kernel directory,
but I could imagine we do more stuff here in the future.

I think this will be generally useful (e.g. ostree and rpm-ostree have
duplicate code for this) as is, but it's specifically prep for using
this in container layer splitting.


jmarrero (approved):
lgtm

PR #262, opened 2022-03-07
Fix clippy lints with `cargo 1.60.0-beta.1 (ea2a21c 2022-02-15)`

cgwalters:
I decided to suppress the ref-vs-not ref one for now, hopefully
clippy gets fixed.


jmarrero (approved):
lgtm

PR #263, opened 2022-03-08
tar/import: Improve error for unsigned commit

cgwalters:
Move the check for object type earlier, so we provide
a clearer error message.  Hit this when trying to
pull an unsigned commit in the tests.


jmarrero (approved):
lgtm

PR #264, opened 2022-03-08
tests: Port two tests to v1 fixture

cgwalters:
Add GPG signature on the commit object to match the current bits.


jmarrero (approved):
lgtm

PR #265, opened 2022-03-08
tests: Two more fixture porting patches

cgwalters:
tests: Port two more tests to v1 fixture

Ongoing work to use the updated fixture.

---

tests: Port another two tests to v1 fixture

Ongoing work to use the updated fixture.

---


jlebon (approved):
Looks sane to me!

PR #268, opened 2022-03-10
tests/fixture: Add concept of "owner"

cgwalters:
Prep for chunk splitting - this is a bit like the dpkg/rpm database.


jmarrero (approved):
LGTM

PR #274, opened 2022-03-30
3 minor patches

cgwalters:
cli: Add more docstrings

Specifically for `commit`.

---

cli: Use `bool::then`

Minor drive by code improvement.

---

tree-wide: Use some captured identifiers in format strings

xref https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings
Another drive by code improvement, no immediate motivation.  I've
just been using it in other places and I think we're fine to hard depend
on Rust 1.58 - it's in RHEL8.6 for example.

---


lib/src/chunking.rs
```diff
@@ -342,7 +342,7 @@ impl Chunking {
                     );
                     Cow::Owned(r)
                 }
-                n => Cow::Owned(format!("{} components", n)),
+                n => Cow::Owned(format!("{n} components")),
```
jlebon, line 345: Oh sweet, will have to remember to start using this more now!

jlebon (approved):

jlebon:
Need to bump MSRV here.

cgwalters:
Updated :new: MSRV

PR #276, opened 2022-04-04
Bump to 0.7.0 (lots of semver-incompat API changes)

cgwalters:
The biggest thing here was the merge of "chunked ostree" support in
https://github.com/ostreedev/ostree-rs-ext/pull/123
cascading into various API changes.

But I think it's working, let's cut 0.7 so the next rpm-ostree can
roll in the support.


cgwalters:
https://github.com/ostreedev/ostree-rs-ext/issues/277 is definitely something that'd require an API break but...eh, we can just cut an 0.8.0 later.

jlebon (approved):

Issue #277, opened 2022-04-04
Fix progress reporting


cgwalters:
OK one ugly thing with current import APIs is that our progress reporting is basically broken:

- We only handle it for unencapsulation, not layered paths
- We don't provide useful information on which layer we're handling
- We operate on the *uncompressed* size, for which we don't know the total
- (related to above) We don't provide the total size

PR #278, opened 2022-04-11
container: Add support for layer fetch notifications

cgwalters:
Part of https://github.com/ostreedev/ostree-rs-ext/issues/277
When we only supported one big ostree tarball, our current
single u64 progress notification was OK.  But clients (including our
CLI) really want more dynamic progress notifications.

This adds support for an optional channel that receives started/finished
notifications for layers.  Note that the client *must* read from
this channel if they request it. The channel is not lossy.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -15,10 +15,12 @@ use std::convert::TryFrom;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use structopt::StructOpt;
+use tokio::sync::mpsc::Receiver;
 use tokio_stream::StreamExt;
 
 use crate::commit::container_commit;
 use crate::container as ostree_container;
+use crate::container::store::{ImportProgress, PreparedImport};
 use crate::container::{Config, ImageReference, OstreeImageReference, UnencapsulateOptions};
 use ostree_container::store::{ImageImporter, PrepareResult};
 use ostree_container::UnencapsulationProgress;
@@ -348,6 +350,46 @@ enum ProgressOrFinish {
     Finished(Result<ostree_container::Import>),
 }
 
+/// Render an import progress notification as a string.
+pub fn layer_progress_format(p: &ImportProgress) -> String {
+    let (starting, s, layer) = match p {
+        ImportProgress::OstreeChunkStarted(v) => (true, "ostree chunk", v),
+        ImportProgress::OstreeChunkCompleted(v) => (false, "ostree chunk", v),
+        ImportProgress::DerivedLayerStarted(v) => (true, "layer", v),
+        ImportProgress::DerivedLayerCompleted(v) => (false, "layer", v),
+    };
+    // podman outputs 12 characters of digest, let's add 7 for `sha256:`.
+    let short_digest = layer.digest().chars().take(12 + 7).collect::<String>();
+    if starting {
+        let size = glib::format_size(layer.size() as u64);
+        format!("Fetching {s} {short_digest} ({size})")
+    } else {
+        format!("Fetched {s} {short_digest}")
+    }
+}
+
+async fn handle_layer_progress_print(mut r: Receiver<ImportProgress>) {
+    while let Some(v) = r.recv().await {
+        println!("{}", layer_progress_format(&v));
+    }
+}
+
+fn print_layer_status(prep: &PreparedImport) {
+    let (stored, to_fetch, to_fetch_size) =
+        prep.all_layers()
+            .fold((0u32, 0u32, 0u64), |(stored, to_fetch, sz), v| {
+                if v.commit.is_some() {
+                    (stored + 1, to_fetch, sz)
+                } else {
+                    (stored, to_fetch + 1, sz + v.size())
+                }
+            });
+    if to_fetch > 0 {
+        let size = crate::glib::format_size(to_fetch_size);
+        println!("layers stored: {stored} needed: {to_fetch} ({size})");
+    }
+}
+
 /// Import a container image with an encapsulated ostree commit.
 async fn container_import(
     repo: &ostree::Repo,
@@ -451,6 +493,7 @@ async fn container_store(
     proxyopts: ContainerProxyOpts,
 ) -> Result<()> {
     let mut imp = ImageImporter::new(repo, imgref, proxyopts.into()).await?;
+    let layer_progress = imp.request_progress();
     let prep = match imp.prepare().await? {
         PrepareResult::AlreadyPresent(c) => {
             println!("No changes in {} => {}", imgref, c.merge_commit);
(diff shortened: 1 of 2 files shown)


lib/src/cli.rs
```diff
@@ -348,6 +350,46 @@ enum ProgressOrFinish {
     Finished(Result<ostree_container::Import>),
 }
 
+/// Render an import progress notification as a string.
+pub fn layer_progress_format(p: &ImportProgress) -> String {
```
cgwalters, line 354: I made this `pub` so we can re-use it in rpm-ostree, which wants to render status via dbus, not `println!`.

cgwalters:
rpm-ostree patch I tested this with too:

```
diff --git a/Cargo.lock b/Cargo.lock
index 074fdcda..5a9db14f 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1625,8 +1625,6 @@ dependencies = [
 [[package]]
 name = "ostree-ext"
 version = "0.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0feb85eda4aa146654a73bad0d2a31282a8e83ca85ffa2ae1a52e696e8df0664"
 dependencies = [
  "anyhow",
  "async-compression",
diff --git a/Cargo.toml b/Cargo.toml
index 5304be5d..70fe5707 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,3 +109,6 @@ bin-unit-tests = []
 sanitizers = []
 
 default = []
+
+[patch.crates-io]
+ostree-ext = { path = "../../ostreedev/ostree-rs-ext/lib" }
diff --git a/rust/src/sysroot_upgrade.rs b/rust/src/sysroot_upgrade.rs
index 9cabb0f3..6022256f 100644
--- a/rust/src/sysroot_upgrade.rs
+++ b/rust/src/sysroot_upgrade.rs
@@ -10,11 +10,12 @@ use ostree_container::store::ImageImporter;
 use ostree_container::store::PrepareResult;
 use ostree_container::OstreeImageReference;
 use ostree_ext::container as ostree_container;
-use ostree_ext::container::store::ManifestLayerState;
+use ostree_ext::container::store::{ImportProgress, ManifestLayerState};
 use ostree_ext::ostree;
 use std::convert::TryFrom;
 use std::pin::Pin;
 use tokio::runtime::Handle;
+use tokio::sync::mpsc::Receiver;
 
 impl From<Box<ostree_container::store::LayeredImageState>> for crate::ffi::ContainerImageState {
     fn from(s: Box<ostree_container::store::LayeredImageState>) -> crate::ffi::ContainerImageState {
@@ -42,16 +43,26 @@ fn layer_counts<'a>(layers: impl Iterator<Item = &'a ManifestLayerState>) -> (u3
     )
 }
 
+async fn layer_progress_print(mut r: Receiver<ImportProgress>) {
+    while let Some(v) = r.recv().await {
+        let msg = ostree_ext::cli::layer_progress_format(&v);
+        output_message(&msg);
+    }
+}
+
 async fn pull_container_async(
     repo: &ostree::Repo,
     imgref: &OstreeImageReference,
 ) -> Result<ContainerImageState> {
     output_message(&format!("Pulling manifest: {}", &imgref));
     let mut imp = ImageImporter::new(repo, imgref, Default::default()).await?;
+    let layer_progress = imp.request_progress();
     let prep = match imp.prepare().await? {
         PrepareResult::AlreadyPresent(r) => return Ok(r.into()),
         PrepareResult::Ready(r) => r,
     };
+    let progress_printer =
+        tokio::task::spawn(async move { layer_progress_print(layer_progress).await });
     let digest = prep.manifest_digest.clone();
     output_message(&format!("Importing: {} (digest: {})", &imgref, &digest));
     let ostree_layers = prep
@@ -72,9 +83,10 @@ async fn pull_container_async(
             "custom layers stored: {stored} needed: {n_to_fetch} ({size})"
         ));
     }
-    let import = imp.import(prep).await?;
+    let import = imp.import(prep).await;
+    let _ = progress_printer.await;
     // TODO log the discarded bits from import
-    Ok(import.into())
+    Ok(import?.into())
 }
 
 /// Import ostree commit in container image using ostree-rs-ext's API.
```

lib/src/cli.rs
```diff
@@ -348,6 +350,46 @@ enum ProgressOrFinish {
...
+async fn handle_layer_progress_print(mut r: Receiver<ImportProgress>) {
+    while let Some(v) = r.recv().await {
+        println!("{}", layer_progress_format(&v));
+    }
+}
+
+fn print_layer_status(prep: &PreparedImport) {
+    let (stored, to_fetch, to_fetch_size) =
+        prep.all_layers()
+            .fold((0u32, 0u32, 0u64), |(stored, to_fetch, sz), v| {
```
lucab, line 380: I think this is fine as-is, but it would be easier to read without the same-names shadowing.

lib/src/container/store.rs
```diff
@@ -286,6 +303,14 @@ impl ImageImporter {
         self.prepare_internal(false).await
     }
 
+    /// Create a channel receiver that will get notifications for layer fetches.
+    pub fn request_progress(&mut self) -> Receiver<ImportProgress> {
+        assert!(self.layer_progress.is_none());
```
lucab, line 308: It sounds like we wanted a spmc channel here instead, but tokio may not have one at hand.

lucab (approved):

lib/src/container/store.rs
cgwalters: I think in practice any "multiple viewers" aspect will need to be handled at a higher level - rpm-ostree has quite complex infrastructure already for this.

lib/src/cli.rs
lucab: I didn't initially realize that you named in the same way (`stored` / `to_fetch`) both the outer results and the inner closure parameters, which tripped me for a second. But I don't think this deserves changing, just note that name clashes are not immediate to spot.

lib/src/cli.rs
```diff
@@ -451,22 +493,20 @@ async fn container_store(
...
-            println!("Using layer: {}", layer.digest());
-        } else {
-            let size = crate::glib::format_size(layer.size());
-            println!("Downloading layer: {} ({})", layer.digest(), size);
-        }
-    }
-    let import = imp.import(prep).await?;
+    print_layer_status(&prep);
+    let progress_printer =
+        tokio::task::spawn(async move { handle_layer_progress_print(layer_progress).await });
```
cgwalters, line 506: Incidentally I think what we want here is https://blog.yoshuawuyts.com/postfix-spawn/ to lifecycle-bind this task to the parent one.  Will look at that at some point.  Maybe https://lib.rs/crates/async_nursery ?

PR #279, opened 2022-04-14
export: Fix `/etc` for chunked archives

cgwalters:
Yeah, the fact that this got past CI shows a big gap.  I am working
on integration tests here that would cover this.
Unit testing is possible, just a little annoying right now because
we need to manually unpack the container image and inspect the tarballs
(can't rely on a container runtime in our unit tests).


lib/src/tar/export.rs
```diff
@@ -541,8 +541,12 @@ pub(crate) fn export_chunk<W: std::io::Write>(
         let (objpath, h) = writer.append_content(checksum.borrow())?;
         for path in paths.iter() {
             let path = path.strip_prefix("/").unwrap_or(path);
+            let path = match path.strip_prefix("usr/etc") {
```
lucab, line 544: This change looks good on its own, but I guess `export_final_chunk()` may need the same kind of path tweaking?

lucab (approved):

lib/src/tar/export.rs
cgwalters: Ooh, you're absolutely right.  I think in practice it wasn't a problem, but yes.

cgwalters:
OK, I discovered more technical debt here while working on cleaning this up, but let's defer that and just get the targeted fix out.

Paths are...confusing.  The ostree C library really wants to operate on absolute paths.  When we're working on a physical filesystem we now prefer using cap-std which wants relative paths, but accepts either case of `./foo` or `foo`.  When serializing to tar, we can use either relative format, but we should be consistent and we're not.

cgwalters:
@lucab (or anyone else) Mind doing a quick re-review this since I updated it?
I did test this though and it still fixes the bug.

lucab (approved):
Earlier I saw the similarity with existing `map_path`, but I didn't want to drag you into that rabbit hole. Thanks for doing the cleanup anyway!

PR #281, opened 2022-04-15
ci: Add a test flow that does privileged integration testing

cgwalters:
Trying to close the gap we have on this repository around
integration testing.  I am hopeful that by merging at least
the ostree{,-rs,-ext} repositories together we will get this
naturally in the future.

This first test crucially also uses the *existing* images as
fixtures, so we'll test compatibility with current ones.

I do still plan to set up OCP Prow for this repo to have reliable
nested virt CI, but...I realized that a pattern that is pretty
useful is that we can run a privileged container that "owns" the
host and execute code there too.

This current CI test actually doesn't run code directly on the
host, it uses ostree's `--sysroot` bits to just write files.

There are interesting aspects to this; for example while Ubuntu doesn't
use SELinux enabled by default, it does
have the kernel code enabled and we can write the `security.selinux`
xattrs just fine from FCOS.

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Assumes that the current environment is a privileged container
+# with the host mounted at /run/host.  We can basically write
+# whatever we want, however we can't actually *reboot* the host.
+set -euo pipefail
+
+sysroot=/run/host
+# Current stable image fixture
+image=quay.io/coreos-assembler/fcos:testing-devel
+# My hand-uploaded chunked images
+chunked_image=quay.io/cgwalters/fcos-chunked:latest
+imgref=ostree-unverified-registry:${image}
+stateroot=testos
+
+set -x
+
+if test '!' -e "${sysroot}/ostree"; then
+    ostree admin init-fs --modern "${sysroot}"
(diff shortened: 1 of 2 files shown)


ci/priv-integration.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Assumes that the current environment is a privileged container
+# with the host mounted at /run/host.  We can basically write
```
lucab, line 3: This clearly works as is, but it isn't very clear to me why this bindmount was introduced. We are only using `ostree` and `ostree-ext-cli`, plus I don't expect the Ubuntu host to have an `/ostree` directory. Is this equivalent to an empty-directory bindmount? Is it to avoid some kind of quota limit?

lucab (approved):

ci/priv-integration.sh
cgwalters: > plus I don't expect the Ubuntu host to have an /ostree directory. 

We're making one here!  That's a nice thing about ostree, we can just install inside an existing system.

cgwalters: Basically, we want to try to install onto the real host filesystem, we don't want to operate on the container `overlayfs`.

PR #282, opened 2022-04-15
container: Precise byte-level progress reporting

cgwalters:
container: Remove current byte-level progress option

We need to rework this to have useful per-layer information.

---

container: Add new byte-level progress

This new byte-level progress API allows clients to render
fine-grained download progress of individual layers.

Notably, it replaces the previous broken API which operated
on the *decompressed* stream - but we don't know the total
size of that.  This one operates on the compressed stream,
and we have the total in the manifest so we can render
an accurate progress bar.

---

cli: Use byte progress for containers

We can now show an accurate download progress.

---

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/277


jmarrero (approved):
lgtm

PR #286, opened 2022-04-27
tests: Consistently use `store::`

cgwalters:
Seeing the full `ostree_ext::container::store::` a lot is too
verbose.  Consistently shorten this to just `store::`.

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -4,7 +4,7 @@ use cap_std::fs::{Dir, DirBuilder};
 use once_cell::sync::Lazy;
 use ostree::cap_std;
 use ostree_ext::chunking::ObjectMetaSized;
-use ostree_ext::container::store::PrepareResult;
+use ostree_ext::container::store;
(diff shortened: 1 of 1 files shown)


lib/tests/it/main.rs
```diff
@@ -4,7 +4,7 @@ use cap_std::fs::{Dir, DirBuilder};
 use once_cell::sync::Lazy;
 use ostree::cap_std;
 use ostree_ext::chunking::ObjectMetaSized;
-use ostree_ext::container::store::PrepareResult;
+use ostree_ext::container::store;
```
lucab, line 7: It looks like you maybe wanted `ostree_ext::container::store::{self, ImageImporter, PrepareResult}`? But this way is ok too.

lucab (approved):

lib/tests/it/main.rs
cgwalters: Hmm...yeah, may consider that in a followup.

PR #290, opened 2022-05-04
build: Be compatible with cap-std-ext 0.25

cgwalters:
We only use the `cmdext` bits right now, not the atomic writes.
Widen our compatibility matrix so we can be built with both.


jmarrero (approved):
lgtm

PR #291, opened 2022-05-06
ci: make use of `cargo-deny` for baseline license check and crate dups

cgwalters:
See https://crates.io/crates/cargo-deny
This looks like a really nice tool, let's use it to validate
license compatibility mainly.   But I also added `sources` and
`bans`.


jmarrero (approved):
pretty cool, lgtm

PR #292, opened 2022-05-06
Fix extra NUL char in IMA xattr, add tests

cgwalters:
I went to add testing for our IMA bits and I tripped over a bug; we
are going out of our way to store a trailing `NUL` character in
the ostree xattrs for the key name.  We should not do this, because
it will break fsck.  What we end up passing to the kernel looks like
`security.ima\0\0`, and the kernel being C will happily take that
to mean `security.ima`.  But the checksum we computed originally is
using `security.ima\0`, not what we will read back from disk as
`security.ima`.

### DIFF
--- a/ci/ima.sh
+++ b/ci/ima.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# Assumes that the current environment is a mutable ostree-container
+# with ostree-ext-cli installed in /usr/bin.  
+# Runs IMA tests.
+set -xeuo pipefail
+
+if test '!' -x /usr/bin/evmctl; then
+    rpm-ostree install ima-evm-utils
+fi
+
+ostree-ext-cli internal-only-for-testing run-ima
+echo ok "ima"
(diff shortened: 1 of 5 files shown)


ci/ima.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Assumes that the current environment is a mutable ostree-container
+# with ostree-ext-cli installed in /usr/bin.  
+# Runs IMA tests.
+set -xeuo pipefail
+
+if test '!' -x /usr/bin/evmctl; then
+    rpm-ostree install ima-evm-utils
```
cgwalters, line 8: It's tempting to add support for IMA using the shared library.  Alternatively, I might argue to just add this FCOS by default.

Or, we could use a separate fixture for this.  Dunno.

jmarrero: If you mean adding `ima-evm-utils` to the FCOS default builds and not just adding it here... I think probably the path of least resistance would be to add a separate fixture.

jmarrero (approved):
lgtm

PR #296, opened 2022-05-11
container: Factor out helper to copy image reference

cgwalters:
Prep for some signing work, but this also may be generally
useful in the future.

---

### DIFF
--- a/lib/src/container/skopeo.rs
+++ b/lib/src/container/skopeo.rs
@@ -1,7 +1,9 @@
 //! Fork skopeo as a subprocess
 
+use super::ImageReference;
 use anyhow::{Context, Result};
 use serde::Deserialize;
+use std::io::Read;
 use std::process::Stdio;
 use tokio::process::Command;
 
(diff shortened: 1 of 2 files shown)


lib/src/container/skopeo.rs
```diff
@@ -55,6 +58,41 @@ pub(crate) fn spawn(mut cmd: Command) -> Result<tokio::process::Child> {
...
+) -> Result<String> {
+    let digestfile = tempfile::NamedTempFile::new()?;
+    let mut cmd = new_cmd();
+    cmd.stdout(std::process::Stdio::null()).arg("copy");
+    cmd.arg("--digestfile");
+    cmd.arg(digestfile.path());
+    cmd.arg(src.to_string());
+    let destarg = tag
+        .map(|t| format!("oci:{destdir}:{t}"))
+        .unwrap_or_else(|| format!("oci:{destdir}"));
```
lucab, line 82: It looks like this was unprefixed before, although I think this new form is the correct one.
Not sure if it was manually checked for unexpected side-effects.

lucab (approved):

lib/src/container/skopeo.rs
cgwalters: OK, now that I look in fact the change here was actually just wrong.  I thought we were testing it in CI, but we're not.

So...let's do https://github.com/ostreedev/ostree-rs-ext/pull/297 first, then I'll fix this code...

cgwalters:
OK this one is just a small fix on top of what was already reviewed, so merging.

PR #297, opened 2022-05-12
tests: Add a case for exporting to ociarchive

cgwalters:
Since this exercises our "build oci, then copy" bits.


jmarrero (approved):
lgtm

PR #299, opened 2022-05-13
import: Fix error message for missing commit metadata

cgwalters:
I'm writing some code around changing commit metadata, and
it took me an embarassingly long amount of time to figure out
that it wasn't my code that was broken - it was this error message.

In this case, output the actual next object type we found instead
of saying "commit" which we already did find.


jmarrero (approved):
lgtm

Issue #295, opened 2022-05-09
Add API and CLI to write detached metadata for a container


cgwalters:
Let's add:
```
$ ostree container write-detached-metadata --from docker://quay.io/exampleos@sha256:e07213cb0393 --replace-file detached --to docker://quay.io/exampleos:latest
```

This would entirely replace the detached metadata embedded in the tar stream with the file `detached`.  We also want higher level sugar like `--inject-gpg-signature gpg.txt` or perhaps even higher level we should have `ostree container sign`.  

Crucially, we should copy all other metadata about the container image (such as labels) again to the target image.

This is in the hot path for switching FCOS to chunked images, xref https://github.com/coreos/fedora-coreos-config/pull/1698#issuecomment-1121473651

PR #301, opened 2022-05-13
Add API and CLI to update detached metadata

cgwalters:
In the FCOS use case and I'm sure others, we want a flow where
we do a build (including a container image) and then once it's ready,
we sign it by passing the commit metadata to a separate system.

Basically what we want is the ability to update the detached metadata
object in an exported container image.

Now, I'm regretting the design choice to have the container flow
reuse the tar path of having the signature be part of the tar stream
instead of part of the container metadata, because it *greatly*
complicates things here, particularly in terms of handling chunked
images.

We want to preserve all metadata and other layers in the image;
we just need to add/replace a single entry in the layer that has
the ostree metadata.

Except, because this ostree layer gets its own special label
in the container image metadata, we need to update that label.

What would make this a lot easier is if we had write support
via skopeo/containers-image-proxy.  Because we don't, given
an image on a remote registry, right now we pull the whole thing
down into a temporary OCI directory, even though we only want to
mutate one layer.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/295


jmarrero (approved):
lgtm

PR #302, opened 2022-05-13
Two signature injection prep patches

cgwalters:
store: Extract a helper function to find ostree layer

Prep for further work around signature injection.

---

tar/export: Extract helper functions

Prep for signature rewriting.

---


jmarrero (approved):
lgtm

PR #306, opened 2022-06-09
lib: Bump to ostree-rs 0.14

cgwalters:
This bumped ostree semver, and since we re-export it also bumps
ours, but we've already got that queued.

When doing this I went to see if there were any new APIs that
we should be using, and saw https://github.com/ostreedev/ostree-rs/pull/50
and was like "ah hah, I remember that now" and a quick grep found
the place I wanted to use it.


jmarrero (approved):
lgtm

PR #310, opened 2022-06-23
container/encapsulate: Drop two `unwrap()`

cgwalters:
I happened to be looking at this code and realized we didn't
need to do either `unwrap()` here.


jmarrero (approved):
lgtm

PR #312, opened 2022-06-24
container/tar: Prep patches for export fixes

cgwalters:
container/tar: Thread commit checksum down into tar writer

Prep for future refactoring where we will want to cache things
related to it.  We don't have any design for emitting multiple
commits into a single tar stream, so drop that implicit ability.

---

container/tar: Pass ownership of chunks down into export

There's no reason to keep all the data alive until the export is
done, and it may allow us to optimize things in the future.

---

tar/export: Make constructor fallible

Prep for a future patch which will add actually fallible code there.

---

tar/export: Load and cache commit object

The writer can only write a single commit.  Having this data
already loaded will help with future patches for the chunked
export.

---

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -107,8 +107,6 @@ impl ObjectMetaSized {
 #[derive(Debug, Default)]
 pub struct Chunking {
     pub(crate) metadata_size: u64,
-    #[allow(dead_code)]
-    pub(crate) commit: Box<str>,
     pub(crate) meta: Vec<Meta>,
     pub(crate) remainder: Chunk,
     pub(crate) chunks: Vec<Chunk>,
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

cgwalters:
Hmm.  The unit tests seem to be hanging/freezing in CI, but I'm not reproducing locally.

cgwalters:
OK: *the plot thickens*.  Something in this PR is making unit tests hang...but so far only in CI :detective:

cgwalters:
Oh oh of course, NOW it works.  Hmmm.

cgwalters:
I'm...going to assume that the VM involved here had gremlins for a bit.  If the hang reoccurs, we can debug further then.

PR #313, opened 2022-06-24
ci: Allow manually invoking tests

cgwalters:
I want to be able to run them manually too to sanity check things.
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#defining-inputs-for-manually-triggered-workflows


jmarrero (approved):
lgtm

PR #317, opened 2022-06-24
container/encapsulate: Pass down options farther

cgwalters:
Pass the `ExportOpts` farther down into processing; prep for
adding more fields and using them there.

Add a helper method that gives us the compression from the options,
instead of open-coding it.


jmarrero (approved):
lgtm

Issue #316, opened 2022-06-24
Build failures when compiling locally


jlebon:
Not sure yet on the difference between my setup and CI, but trying to build locally I'm hitting

```
   Compiling ostree-ext v0.8.0 (/var/home/jlebon/Code/github.com/ostreedev/ostree-rs-ext/lib)
error[E0599]: no method named `send_replace` found for reference `&tokio::sync::watch::Sender<std::option::Option<LayerProgress>>` in the current scope
   --> lib/src/container/unencapsulate.rs:211:26
    |
211 |                 progress.send_replace(Some(status));
    |                          ^^^^^^^^^^^^ method not found in `&tokio::sync::watch::Sender<std::option::Option<LayerProgress>>`

error[E0599]: no method named `atomic_write` found for reference `&ostree::cap_std::fs::Dir` in the current scope
   --> lib/src/container/ocidir.rs:120:13
    |
120 |         dir.atomic_write("oci-layout", r#"{"imageLayoutVersion":"1.0.0"}"#)?;
    |             ^^^^^^^^^^^^ method not found in `&ostree::cap_std::fs::Dir`

error[E0599]: no method named `atomic_replace_with` found for struct `Arc<ostree::cap_std::fs::Dir>` in the current scope
   --> lib/src/container/ocidir.rs:134:18
    |
134 |                 .atomic_replace_with(&path, |w| std::io::copy(&mut src, w))?;
    |                  ^^^^^^^^^^^^^^^^^^^ method not found in `Arc<ostree::cap_std::fs::Dir>`

error[E0599]: no method named `atomic_replace_with` found for struct `Arc<ostree::cap_std::fs::Dir>` in the current scope
   --> lib/src/container/ocidir.rs:262:14
    |
262 |             .atomic_replace_with("index.json", |w| -> Result<()> {
    |              ^^^^^^^^^^^^^^^^^^^ method not found in `Arc<ostree::cap_std::fs::Dir>`

```

```
$ rustc --version
rustc 1.61.0 (fe5b13d68 2022-05-18)
```

PR #320, opened 2022-06-24
lib: Bump tokio minimal version to 1.13.0

jlebon:
We use `Sender::send_replace` which was added in 1.13.0:
https://github.com/tokio-rs/tokio/pull/3962

Closes: #316

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -17,7 +17,7 @@ bitflags = "1"
 camino = "1.0.4"
 chrono = "0.4.19"
 cjson = "0.1.1"
-cap-std-ext = ">= 0.24, <= 0.25"
+cap-std-ext = ">= 0.25"
 cap-tempfile = "0.24"
 flate2 = { features = ["zlib"], default_features = false, version = "1.0.20" }
 fn-error-context = "0.2.0"
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -39,7 +39,7 @@ structopt = "0.3.21"
 tar = "0.4.38"
 tempfile = "3.2.0"
 term_size = "0.3.2"
-tokio = { features = ["full"], version = "1" }
+tokio = { features = ["full"], version = ">= 1.13.0" }
 tokio-util = { features = ["io-util"], version = "0.6.9" }
 tokio-stream = { features = ["sync"], version = "0.1.8" }
 tracing = "0.1"


jlebon:
Wait, looks like there's more undeclared requirements.

jlebon:
OK, got a local build now! I'm not quite sure why CI isn't hitting this. Probably related to the use of Swatinem/rust-cache?

lib/Cargo.toml
```diff
@@ -17,7 +17,7 @@ bitflags = "1"
 camino = "1.0.4"
 chrono = "0.4.19"
 cjson = "0.1.1"
-cap-std-ext = ">= 0.24, <= 0.25"
+cap-std-ext = ">= 0.25"
```
cgwalters, line 20: I have a pending PR for this I haven't published, but I will; it doesn't look related to this PR though?

cgwalters: Oh nevermind right,
```
$ cargo update -p cap-std-ext --precise 0.24.3
$ cargo check
error[E0599]: no method named `atomic_write` found for reference `&ostree::cap_std::fs::Dir` in the current scope
   --> lib/src/container/ocidir.rs:120:13
    |
120 |         dir.atomic_write("oci-layout", r#"{"imageLayoutVersion":"1.0.0"}"#)?;
    |             ^^^^^^^^^^^^ method not found in `&ostree::cap_std::fs::Dir`
```
r

cgwalters:
Isn't the problem most likely that you have a local build for which you haven't run `cargo update` in a while?

lib/Cargo.toml
jlebon: It's related in the sense that it's also required for me to get a successful build locally. :) It's in a separate commit:

```
lib: Bump cap-std-ext minimal version to 0.25

We use `atomic_write` which was added in 0.25:
https://github.com/coreos/cap-std-ext/pull/12
```

jlebon:
> Isn't the problem most likely that you have a local build for which you haven't run `cargo update` in a while?

Oh yeah, that's clearly the immediate cause. I think it's just been a while since I've hacked on a Rust library (since for binaries we have Dependabot hooked up).

cgwalters:
OK fun, this one also hit https://github.com/ostreedev/ostree-rs-ext/issues/322

cgwalters (approved):

PR #321, opened 2022-06-24
Use single-threaded tokio by default

cgwalters:
I saw https://www.reddit.com/r/rust/comments/v8e9fa/local_async_executors_and_why_they_should_be_the/
go by and I think it's a good argument why one should use single-threaded
tokio by default.

There's more to do around potentially using
https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html
etc.


cgwalters:
Oh oh right I see, this actually isn't really tested until we *also* have https://github.com/containers/containers-image-proxy-rs/pull/32

jmarrero (approved):
lgtm

PR #323, opened 2022-06-24
integrationtest: Properly thread `async` through to test helper

cgwalters:
In various places we're lazy about properly using async; e.g.
some of the test fixtures do a ton of blocking I/O.  None of that
really matters honestly as long as the *core* library does
things properly.

This bit of code in the test case was very hackily re-entering
the async context that we already have from `main`; but it only
works on the multithreaded tokio and we're aiming to switch to
single threaded.

That was just lazy of me; it was actually not at all hard to just
properly add a few `async fn` and `await`.


jmarrero (approved):
lgtm

PR #324, opened 2022-06-24
Port to cap-std 0.25, ostree 0.15, sh-inline 0.3

cgwalters:
Our semver train is fun; updating cap-std for us involves a semver
bump for ostree because it exports public types from that.

We also have our utility crate `cap-std-ext` in the mix that needs
bumping.  And we have `sh-inline` which depends on all this.

I had to bump semver for all those at once.  I'm now regretting
`sh-inline`'s existence, or at least its dependency on cap-std.

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -17,20 +17,20 @@ bitflags = "1"
 camino = "1.0.4"
 chrono = "0.4.19"
 cjson = "0.1.1"
-cap-std-ext = ">= 0.25"
-cap-tempfile = "0.24"
+cap-std-ext = "0.26"
+cap-tempfile = "0.25"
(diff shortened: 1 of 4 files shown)


cgwalters:
OK yeah this all turned into a huge mess; my little on the side crate required https://github.com/cgwalters/rust-sh-inline/commit/ea72940ab27fff72943d24921b26137178f59024
and in practice I think its semver kind of broke with the cap-std-ext release since I badly claimed in https://github.com/cgwalters/rust-sh-inline/commit/9a86259ae77111b83242324813e841385f6efd21 but that turned out to be a lie.

Anyways...sorted out here for now but...hopefully cap-std gets merged into std officially, which would put an end to both semver bumps there *and* the awkward dependency chains.

jmarrero (approved):
:shipit:

PR #325, opened 2022-06-26
container: Enable compression by default

cgwalters:
I'd noticed that some of our images were large but hadn't looked
closely.  I now really wish I had...

I think this regressed quite a while ago during a refactoring.

### DIFF
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -237,7 +237,7 @@ async fn build_impl(
 ) -> Result<String> {
     let mut opts = opts.unwrap_or_default();
     if dest.transport == Transport::ContainerStorage {
-        opts.compress = false;
+        opts.skip_compression = false;
     }
     let digest = if dest.transport == Transport::OciDir {
         let _copied: ImageReference = build_oci(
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
Note I plan to circle back here and use the infrastructure in https://github.com/ostreedev/ostree-rs-ext/pull/327 to add a unit test for this.

lib/src/container/encapsulate.rs
```diff
@@ -283,8 +283,8 @@ async fn build_impl(
 /// Options controlling commit export into OCI
 #[derive(Debug, Default)]
 pub struct ExportOpts {
-    /// If true, perform gzip compression of the tar layers.
-    pub compress: bool,
+    /// If false, do not perform gzip compression of the tar layers.
+    pub skip_compression: bool,
```
jlebon, line 287: This looks odd. Shouldn't it be "If false, perform gzip compression of the tar layers" ? (Or maybe reword as "If true, skip gzip compression of the tar layers".) Another way to see this I think is we added one negation in the field name, but two negations in the docstring.

lib/src/container/encapsulate.rs
```diff
@@ -237,7 +237,7 @@ async fn build_impl(
 ) -> Result<String> {
     let mut opts = opts.unwrap_or_default();
     if dest.transport == Transport::ContainerStorage {
-        opts.compress = false;
+        opts.skip_compression = false;
```
jlebon, line 240: And this should be `true`, no?

cgwalters:
Thanks, followup in https://github.com/ostreedev/ostree-rs-ext/pull/330

PR #326, opened 2022-06-27
tests: Deduplicate tar structure validation code

cgwalters:
Most of the expected entries are common between v0 and v1.
Prep for some further code refactoring.


jmarrero (approved):
lgtm

Issue #309, opened 2022-06-23
chunked: Fix `/tmp` and `/var`


cgwalters:
So embarassingly, this fails right now:

```Dockerfile
FROM quay.io/coreos-assembler/fcos:rawhide
RUN rpm-ostree install cowsay
```

```
$ podman build -t localhost/fcos-cowsay .
STEP 1/2: FROM quay.io/coreos-assembler/fcos:rawhide
STEP 2/2: RUN rpm-ostree install cowsay
error: No such file or directory (os error 2) at path "/tmp/.tmpSc0EVK"
$
```

because we're missing `/tmp` in chunked images.

PR #328, opened 2022-06-27
chunked: Include all directories

cgwalters:
Depends:

- https://github.com/ostreedev/ostree-rs-ext/pull/327
- https://github.com/ostreedev/ostree-rs-ext/pull/326

---

tests: Validate structure of chunked image ostree layer

Chunked images are format v1 tar; let's add some unit-test style
code that validates the tar structure there in the same way we
verify direct tar exports.

Prep for fixing https://github.com/ostreedev/ostree-rs-ext/issues/309

---

chunking: Include all directories in ostree layer

This is prep for fixing https://github.com/ostreedev/ostree-rs-ext/issues/309
but it will not fix it on its own; to do that we need to introduce
a new chunked format.  That is forthcoming.

In the default tar export, we walk a commit from the "root".  When
exporting chunked images (what will soon be called "chunkedv0")
we instead export content objects, then a final layer which
contains the ostree metadata (commit and dirtree/dirmeta).

But in that "chunkedv0" flow, nothing creates the *tar* equivalent
of directories.  This has a number of problems; primarily it will
implicitly omit any directories which do not have content in them
(for example `/tmp`).  We also won't have the proper ownership/directories
reflected in the tar stream.

In this commit, change how we generate the "ostree layer" which
contains the ostree metadata to walk from the ostree-commit
as a root; except we just don't emit any content objects in that
flow to start, assuming they were emitted earlier.

(It is common for chunkedv0 images to have content objects, they
 are just emitted as a second phase)

Note that we no longer need to gather all metadata objects when
processing chunking, because we'll end up walking everything in
the export phase instead.

---

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -8,6 +8,7 @@ use ostree_ext::container::store;
 use ostree_ext::container::{
     Config, ExportOpts, ImageReference, OstreeImageReference, SignatureSource, Transport,
 };
+use ostree_ext::ocidir;
 use ostree_ext::prelude::FileExt;
 use ostree_ext::tar::TarImportOptions;
 use ostree_ext::{gio, glib};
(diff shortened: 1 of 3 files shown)


lib/tests/it/main.rs
```diff
@@ -306,7 +306,10 @@ fn validate_tar_v1<R: std::io::Read>(mut src: tar::Archive<R>) -> Result<()> {
     .into_iter()
     .map(Into::into);
 
-    let expected = prelude.chain(common_tar_structure());
+    let content = [("usr", Directory, 0o755), ("boot", Directory, 0o755)];
```
cgwalters, line 309: I also did cross-verify that this test fails when we don't include the corresponding changes to `export.rs`, and I just looked at the tarballs generated by `cargo run -- internal-only-for-testing create-fixture` before and after.

jmarrero (approved):
lgtm

PR #329, opened 2022-06-27
tar/export: Drop a leftover `dbg!`

cgwalters:
tar/export: Drop a leftover `dbg!`

Oops.

---

ci: Add a custom lint check for leftover `dbg!`

It's tempting to add this to clippy.

---


jmarrero (approved):
lgtm

PR #330, opened 2022-06-27
Compress fixes

cgwalters:
container: Do continue to skip compression for `containers-storage:`

Fixes previous commit.

---

container/encapsulate: Fix `skip_compression` docstring

Remove an accidental double negative.

---


jlebon:
This doesn't not LGTM!

PR #331, opened 2022-06-27
container: Introduce an `ExportLayout::ChunkedV1`

cgwalters:
Depends: https://github.com/ostreedev/ostree-rs-ext/pull/328

---
container: Introduce an `ExportLayout::ChunkedV1`

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/309

I discovered that (what are now called) "chunkedv0" images were missing some
directories like `/tmp`.  This is because the way chunking was implemented
was...basically broken.

In chunking v0 we have:

- content objects
- content objects
- ...
- ostree layer (commit, ostree metadata)
- Any derived layers

But...we really want to "mirror" in the tar stream in a proper
fashion everything that's in the ostree layer. Particularly, in order
to ensure e.g. correct permissions/ownership, the directory entries
must come first.

And logically, it makes sense to have the layer bearing the ostree
commit and the ostree metadata come first.

So the new "ChunkedV1" format is:

- ostree layer (commit, ostree metadata, all directories_
- content objects
- content objects
- ...
- Any derived layers

The ChunkedV1 format can be identified by a new image label:
`const OSTREE_FINAL_LAYER_LABEL: &str = "ostree.final-diffid";`
this label points to the last content object layer.

Implementation wise, this is mostly adding conditionals in
various places.

I'm perhaps being very conservative here in *also* continuing to
support 'chunkedv0 images.  Right now they're not deployed
widely (AFAIK), just shipping Fedora Rawhide this way.

But on the plus side, we'll be able to kill off all the old format
code when this stuff is all stable.

---

container: Condense `ExportLayout` to just V0 and V1

I had a galaxy brain moment earlier today when looking at this PR;
a lot of the matching was doing `SingleLayer | ChunkedV1`.  Everything
just gets simpler if we consider the "single layer" a special case
of a "chunked" image with a single chunk - which we already support!

In the export code, if no content mapping is provided we create
a default chunking.  (This can and should be optimized later to
avoid traversing all objects up front; we can special case this)

Another way to look at this is that the change in the export side
from the tar stream perspective is that we now have:

- all directories and ostree metadata
- all content objects in hash ordering

And here's a key bit: The old (<= ostree-ext 0.7) tar parser
will still parse this just fine - it doesn't currently care about
object ordering.  And the same is true for container runtimes.

---

container: Make `export_layout` bit in prepared import `pub`

Right now we're not using it, but I think we should support
callers e.g. logging it at least.

---

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -10,6 +10,7 @@ use crate::refescape;
 use anyhow::{anyhow, Context};
 use containers_image_proxy::{ImageProxy, OpenedImage};
 use fn_error_context::context;
+use futures_util::TryFutureExt;
 use oci_spec::image::{self as oci_image, Descriptor, History, ImageConfiguration, ImageManifest};
 use ostree::prelude::{Cast, ToVariant};
 use ostree::{gio, glib};
@@ -170,6 +171,9 @@ impl ManifestLayerState {
 /// Information about which layers need to be downloaded.
 #[derive(Debug)]
 pub struct PreparedImport {
+    /// The format we found from metadata
+    #[allow(dead_code)]
+    pub(crate) export_layout: ExportLayout,
     /// The manifest digest that was found
     pub manifest_digest: String,
     /// The deserialized manifest.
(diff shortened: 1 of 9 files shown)


cgwalters:
I've pushed a fixture for this PR to `quay.io/cgwalters/fcos-chunked:v1`.  I also verified that we now have the expected directories like `/tmp` and `/boot`, and that `rpm-ostree install cowsay` works in this container.

lib/src/container/store.rs
```diff
@@ -170,6 +171,9 @@ impl ManifestLayerState {
 /// Information about which layers need to be downloaded.
 #[derive(Debug)]
 pub struct PreparedImport {
+    /// The format we found from metadata
+    #[allow(dead_code)]
```
jmarrero, line 175: This is here just as prep for a next patch?

jmarrero (approved):
lgtm generally just have a question on why we need: https://github.com/ostreedev/ostree-rs-ext/pull/331/files#r908499716 if it's being used on the module.

cgwalters:
OK I pushed another commit on top here `container: Condense ExportLayout to just V0 and V1` which simplifies things a lot.  More cleanup can happen there even too.

I didn't yet cross-verify that the *old* ostree-ext code parses the new tar serialization order, but doing that now.

cgwalters:
OK, test coverage in https://github.com/ostreedev/ostree-rs-ext/pull/335

PR #333, opened 2022-06-28
lib/container: `#[allow(dead_code)]` for ocidir

cgwalters:
When compiling from an external crate, dead code analysis hits
a few things that we don't currently use.


cgwalters:
(For some reason the default for PRs here seems to have switched to draft...maybe just for me?)

jmarrero:
> (For some reason the default for PRs here seems to have switched to draft...maybe just for me?)

I think it might just use whatever one uses last.

jmarrero (approved):
lgtm

PR #335, opened 2022-06-28
tests: Cover old ostree parsing latest code

cgwalters:
In a recent update we changed the serialization format just
for chunked images; we need to test that the *old* ostree knows how to
parse at least our changes to the v0 format, which should be compatible.

Today rpm-ostree vendors a stable ostree-rs-ext; this test leverages
that.


jmarrero (approved):
lgtm

PR #338, opened 2022-07-18
tests: Prep for fixing identical subdirectory export

cgwalters:
tests: Emit expected modes in octal

Easier to read this way.

---

tests: Add two subdirectories with identical content

Prep for fixing https://github.com/coreos/fedora-coreos-tracker/issues/1258

---

tests: Pass tar entries by reference

Prep for testing the full tar export vs chunked container paths
better.

---

tests: More refactoring of expected tar handling

Add a missing 'return Ok(())` so we can split the tar stream validation
to do:

- validate prelude
- validate contents

We were also missing validation for content objects before, so fix
that!

The chunked container path in particular will need to handle content
validation differently from the "all in one" tarball.

---

tests: Verify at least one pkgdb entry is there

Prep for fixing https://github.com/ostreedev/ostree-rs-ext/issues/339

---

### DIFF
--- a/lib/src/fixture.rs
+++ b/lib/src/fixture.rs
@@ -131,6 +131,8 @@ static OWNERS: Lazy<Vec<(Regex, &str)>> = Lazy::new(|| {
         ("usr/bin/hardlink.*", "testlink"),
         ("usr/etc/someconfig.conf", "someconfig"),
         ("usr/etc/polkit.conf", "a-polkit-config"),
+        ("usr/lib/pkgdb", "pkgdb"),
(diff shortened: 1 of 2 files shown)


cgwalters:
deny failure should be fixed by https://github.com/ostreedev/ostree-rs-ext/pull/340

cgwalters:
All of this is refactoring and cleanups to the test suite in preparation for the bugfix to the actual export logic that lives in https://github.com/ostreedev/ostree-rs-ext/pull/341

jlebon (approved):
I'm not familiar with this code, but it seems sane.

Issue #339, opened 2022-07-18
chunked: do export identical subdirectories


cgwalters:
See https://github.com/coreos/fedora-coreos-tracker/issues/1258

Regression from https://github.com/ostreedev/ostree-rs-ext/pull/331

Currently rpm-ostree emits two identical subdirectories in `/usr/lib/sysimage/rpm-ostree-base-db` and `/usr/share/rpm`, and the chunking export skips emitting this incorrectly.

PR #341, opened 2022-07-18
chunking: Recurse on duplicate directory trees

cgwalters:
See https://github.com/coreos/fedora-coreos-tracker/issues/1258

Regression from https://github.com/ostreedev/ostree-rs-ext/pull/331

Currently rpm-ostree emits two identical subdirectories in
`/usr/lib/sysimage/rpm-ostree-base-db` and `/usr/share/rpm`, and the
chunking export skips emitting this incorrectly.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/339

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -120,15 +120,15 @@ fn push_dirtree(
     repo: &ostree::Repo,
     gen: &mut Generation,
     checksum: &str,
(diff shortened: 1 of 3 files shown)


cgwalters:
Depends https://github.com/ostreedev/ostree-rs-ext/pull/338

jlebon (approved):
Seems sane to me.

PR #345, opened 2022-07-19
lib: port CLI handling to clap v3

cgwalters:
This reworks CLI handling logic, porting to latest clap v3.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/373

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -10,8 +10,7 @@ readme = "README.md"
 [dependencies]
 anyhow = "1.0"
 ostree-ext = { path = "../lib" }
-clap = "2.33.3"
-structopt = "0.3.21"
+clap = "3.2"
 libc = "0.2.92"
 tokio = { version = "1", features = ["macros"] }
 log = "0.4.0"
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -11,12 +11,12 @@ version = "0.8.7"
 [dependencies]
 anyhow = "1.0"
 containers-image-proxy = "0.5.1"
-
 async-compression = { version = "0.3", features = ["gzip", "tokio"] }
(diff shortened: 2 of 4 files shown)


lucab:
I'm doing some other clap-v3 porting, so I can take this over too.

lucab:
For a start, I rebased the branch and brought it to green CI state.
There are tons of (deprecation) warnings right now though, I'd like to sort them out here.

lucab:
Ok, porting here is completed and this is ready for review.
Please note that this is an API breaking change.
Also, `clap` v4 just got released and it is going to be another API break. We may want to consider switching to that before tagging a release here.

cgwalters:
Yeah in some conversation over at https://www.reddit.com/r/rust/comments/xqiqfy/clap_400_a_rust_argument_parser_is_released/ it seems like some people were aware a v4 was coming and held off porting from v2.  But, I guess since we're doing this we might as well do v4?

Potentially related to this, we could split off the CLI code into a separate crate with its own semver churn.  Definitely some cost/benefit to that.

Anyways...yeah I think I'd say we should go v4, but I haven't tried the port either myself.  I can do it if you like.

lucab:
@cgwalters when we land this, the clap-v4 update should be just a Cargo manifest bump (as this PR already gets rid of all the planned deprecations).

cgwalters:
LGTM as is but I can't approve this as it's my own PR so...feel free to approve :smile:

cgwalters:
> Please note that this is an API breaking change.

Oh, right.  Well, we forgot to bump semver for this.  Urgh.  Well...ok.  In practice I can't imagine anything is using this API except rpm-ostree. 

If something is, well, I can't imagine we're causing them that much pain.

PR #348, opened 2022-08-02
lib: Update to indicatif 0.17

cgwalters:
Lots of improvements:
https://github.com/console-rs/indicatif/releases/tag/0.17.0

We'll be doing the matching change in rpm-ostree too so it's
good to get down to one vendored version.


ashcrow:
cc @lucab

lucab (approved):

PR #349, opened 2022-08-02
cli: Add `container encapsulate --compression-fast`

cgwalters:
container: Make `skip_compression` really use gzip fast, not none

Not doing any compression at all is quite bad for tar streams
which have a ton of duplicate data.

---

cli: Add `container encapsulate --compression-fast`

For the use case of exporting to a temporary OCI directory, which
we in turn then want to push into `containers-storage:` which
is going to decompress it again.

This will be used for e.g.
https://github.com/coreos/rpm-ostree/issues/3900

---

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -131,6 +131,10 @@ enum ContainerOpts {
         /// Corresponds to the Dockerfile `CMD` instruction.
         #[structopt(long)]
         cmd: Option<Vec<String>>,
+
+        /// Compress at the fastest level (e.g. gzip level 1)
+        #[structopt(long)]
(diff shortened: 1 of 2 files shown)


ashcrow:
cc @lucab

lucab:
This short-circuiting of `skip_compression` to `Compression::fast` seems like a footgun, I hope it won't last long (I'd rather just rename it now).
It seems like the intent is to always mandate some kind of compression, so perhaps a better switch would be `--compression=fast|medium|best`.

That said, if this is urgently needed to unblock CI, feel free to merge it as is.

lucab (approved):

PR #351, opened 2022-08-17
container: write `base` ref

jlebon:
Although the OSTree commit checksum is in a label on the container
image, sometimes we need to be able to inspect the commit from within
the container itself.

Rather than forcing tools to look through the whole repo for a commit
object, write a canonical `base` ref pointing to it.

Related: https://github.com/ostreedev/ostree/pull/2691

### DIFF
--- a/lib/src/tar/export.rs
+++ b/lib/src/tar/export.rs
@@ -26,6 +26,8 @@ pub const FORMAT_VERSIONS: RangeInclusive<u32> = 0..=1;
 const SYSROOT: &str = "sysroot";
 // This way the default ostree -> sysroot/ostree symlink works.
 const OSTREEDIR: &str = "sysroot/ostree";
+// The ref added (under ostree/) in the exported OSTree repo pointing at the commit.
+const OSTREEREF: &str = "encapsulated";
 
 /// In v0 format, we use this relative path prefix.  I think I chose this by looking
 /// at the current Fedora base image tar stream.  However, several others don't do
@@ -320,6 +322,13 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
                 &commitmeta,
             )?;
         }
+
+        // and add the canonical ref
+        let path: Utf8PathBuf = format!("{}/repo/refs/heads/ostree", OSTREEDIR).into();
+        self.append_default_dir(&path)?;
+        let path: Utf8PathBuf =
+            format!("{}/repo/refs/heads/ostree/{}", OSTREEDIR, OSTREEREF).into();
+        self.append_default_data(Utf8Path::new(&path), self.commit_checksum.as_bytes())?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


lib/src/tar/export.rs
```diff
@@ -26,6 +26,8 @@ pub const FORMAT_VERSIONS: RangeInclusive<u32> = 0..=1;
 const SYSROOT: &str = "sysroot";
 // This way the default ostree -> sysroot/ostree symlink works.
 const OSTREEDIR: &str = "sysroot/ostree";
+// The ref added in the exported OSTree repo pointing at the base commit.
+const BASEREF: &str = "base";
```
cgwalters, line 30: Let's bikeshed this a bit...

 - `container-encapsulated-commit`
 - `default`
 - `ostree/container/encapsulated`
 - `ostree/default`

I think I'd at least vote to prefix this with `ostree/` to be clear it's an invented ostree thing.

jlebon: I went with `ostree/encapsulated`. `commit` and `container` felt redundant, and `default` too generic. I like that `encapsulated` highlights that it's purely an encapsulation concept and not something we'd write to disk on import.

cgwalters: Cool, SGTM.

cgwalters (approved):

lib/src/tar/export.rs
```diff
@@ -26,6 +26,8 @@ pub const FORMAT_VERSIONS: RangeInclusive<u32> = 0..=1;
 const SYSROOT: &str = "sysroot";
 // This way the default ostree -> sysroot/ostree symlink works.
 const OSTREEDIR: &str = "sysroot/ostree";
+// The ref added (under ostree/) in the exported OSTree repo pointing at the commit.
+const OSTREEREF: &str = "encapsulated";
```
cgwalters, line 30: I think this should be `ostree/encapsulated` though; that's what the ref is.  It's the string you'd pass to `ostree rev-parse`.

lib/src/tar/export.rs
```diff
@@ -320,6 +322,13 @@ impl<'a, W: std::io::Write> OstreeTarWriter<'a, W> {
                 &commitmeta,
             )?;
         }
+
+        // and add the canonical ref
+        let path: Utf8PathBuf = format!("{}/repo/refs/heads/ostree", OSTREEDIR).into();
```
cgwalters, line 327: Then here drop the `/ostree`.

cgwalters: (Also, this could be spelled `Utf8Path::new(OSTREEDIR).join("repo/refs/heads")`)

jlebon: > I think this should be `ostree/encapsulated` though; that's what the ref is. It's the string you'd pass to `ostree rev-parse`.
>
> Then here drop the `/ostree`.

That was my initial plan, but to have this code be generic over all values of `OSTREEREF`, I didn't think it was worth the added complexity of iterating over all the components under `heads/` of the ref and emitting tar entries for every subdir in the path. I *think* most extractors would correctly handle them missing, but it doesn't feel proper skipping them.

> (Also, this could be spelled `Utf8Path::new(OSTREEDIR).join("repo/refs/heads")`)

Hmm, the `format!(...).into()` pattern seems to be prevalent in this file. Should I change those too?

cgwalters: > Hmm, the format!(...).into() pattern seems to be prevalent in this file. Should I change those too?

Nah it's fine, we can do that later.

PR #353, opened 2022-08-22
container: Add an `unencapsulate-self` API and CLI

cgwalters:
This is basically a thin wrapper for
````
rev=$(ostree rev-parse ostree/encapsulated)
ostree pull-local
```

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -98,6 +98,19 @@ enum ContainerOpts {
         quiet: bool,
     },
 
+    #[structopt(alias = "import")]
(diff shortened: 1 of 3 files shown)


cgwalters:
I wrote this, then...now that I'm done I'm not sure it's really worth carrying versus just inlining the above.  I'd thought we need to care about older encapsulated containers but we probably don't, we can just get the updated ostree-ext and rpm-ostree shipped pretty quickly and ignore the legacy case of not having the `ostree/encapsulated` ref.

PR #354, opened 2022-08-22
Release 0.8.3

cgwalters:
This just includes a change to generate an `ostree/encapsulated`
ref in containers, and a new API to fetch both manifest and config.

---


cgwalters:
OK rebased to now pick up the API addition too.

jlebon (approved):

cgwalters:
I published this, and now it's yanked because https://github.com/ostreedev/ostree-rs-ext/pull/355

PR #355, opened 2022-08-23
Revert writing `ostree/encapsulated` ref

cgwalters:
This partially reverts 30dee81c22ad5cb90e77198d3ddbcc25d388afb5
We *do* keep the code which ignores the presence of the file during
import, as an aid to future compatibility.

We can't do this by default for quite a while until support
for reading/ignoring the file has been out in the wild for e.g.
at least a month or two.


cgwalters:
See https://github.com/coreos/rpm-ostree/pull/3957#issuecomment-1224381546

jlebon (approved):
OK fun, thanks for that! Should we track reverting the revert somewhere so we remember to re-add it?

cgwalters:
Perhaps we should make it opt-in...something like a new `--format-version 1.1` flag and API?

PR #356, opened 2022-08-23
 container: Hard require `ostree.bootable` key

cgwalters:
tests: Propagate metadata from prior commit

We want in particular things like `ostree.bootable`.

---

tests: Use `#[track_caller]`

This way panics are properly attributed to the real source.

---

tests/fixture: Inject `ostree.bootable` by default

Since that's our intended use case for now.

---

container: Hard require `ostree.bootable` key

We want to give a clear and useful error when someone tries
to pull a non-ostree-based container image.

---

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -309,6 +309,11 @@ pub(crate) fn parse_manifest_layout<'a>(
     Vec<&'a Descriptor>,
 )> {
     let config_labels = config.config().as_ref().and_then(|c| c.labels().as_ref());
+    let bootable_key = *ostree::METADATA_KEY_BOOTABLE;
+    let bootable = config_labels.map_or(false, |l| l.contains_key(bootable_key));
+    if !bootable {
+        anyhow::bail!("Target image does not have {bootable_key} label");
+    }
 
     let first_layer = manifest
         .layers()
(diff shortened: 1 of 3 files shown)


lib/src/container/store.rs
```diff
@@ -309,6 +309,11 @@ pub(crate) fn parse_manifest_layout<'a>(
     Vec<&'a Descriptor>,
 )> {
     let config_labels = config.config().as_ref().and_then(|c| c.labels().as_ref());
+    let bootable_key = *ostree::METADATA_KEY_BOOTABLE;
+    let bootable = config_labels.map_or(false, |l| l.contains_key(bootable_key));
+    if !bootable {
+        anyhow::bail!("Target container does not have {bootable_key} label");
```
jlebon, line 315: Minor/optional: `s/container/image/`

jlebon (approved):

cgwalters:
I was worried this would break more things, but I did double check we had this key in FCOS and the Fedora Silverblue builds, etc.  And that's the case since rpm-ostree does it by default.

But the test suite needed https://github.com/coreos/rpm-ostree/pull/3971

And clearly we should have `--copy-all-meta` or `--inherit` or something for `ostree commit`.

PR #358, opened 2022-08-25
tests: Use API instead of forking off ostree via bash

cgwalters:
This is just cleaner.  Or it WOULD be if there wasn't a bug in
`OstreeRepoFile` we need to work around.


cgwalters:
Ended up accidentally squashing this in https://github.com/ostreedev/ostree-rs-ext/pull/359 - might as well do it as one unit since that had a few other tweaks too.

Issue #273, opened 2022-03-28
container: support whiteouts


cgwalters:
The use case for removing stuff is in general IMO pretty weak, but it's going to confuse people and break things until we honor whiteouts on the container side.

e.g.
```
RUN touch /etc/foo
RUN rm /etc/foo
```

Or just e.g.:
```
RUN rm /usr/lib/systemd/system/foo.service
```

cgwalters:
The lack of this is actually breaking https://github.com/coreos/coreos-layering-examples/pull/31/commits/3118cafdc6911b283e081ea9d082b9ac26646150

```
[core@cosa-devsh ~]$ ls -al /usr/lib/modules
total 12
drwxr-xr-x.  4 root root  118 Jan  1  1970 .
drwxr-xr-x. 38 root root 4096 Jan  1  1970 ..
drwxr-xr-x.  7 root root 4096 Jan  1  1970 4.18.0-372.19.1.el8_6.x86_64
drwxr-xr-x.  7 root root 4096 Jan  1  1970 4.18.0-372.24.1.el8_6.x86_64
----------.  1 root root    0 Aug 25 00:37 .wh.4.18.0-372.19.1.el8_6.x86_64
```

PR #359, opened 2022-08-25
container/store: Process whiteouts

cgwalters:
A long time ago, in a galaxy far far away, a few of us in the
ostree space did some initial work on Docker-related things; specifically
support for whiteouts landed in
https://github.com/ostreedev/ostree/commit/baaf7450da8a3870e8a42f0cdd4e0ea0ed5018d6

I wish at that time we'd realized how we could more natively
support fetching containers; but, it never occurred to me to fork
off skopeo to do all the heavy lifting for the *fetch* side, which
would have been a lot of work to reimplement particularly in C.
Oh well, better late than never!

Anyways, that whiteout processing was only designed to happen at
checkout time - i.e. when materializing the final filesystem tree.  I
think this was actually a misdesign and we should add
`ostree_mutable_tree_write_with_whiteouts` so that the whiteouts
are processed in-memory.

However for now, there's a relatively low cost to temporarily
materializing the merged tree via hardlinks and handle whiteouts
via the existing code, so let's do that.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/273

---

### DIFF
--- a/lib/src/ostree_manual.rs
+++ b/lib/src/ostree_manual.rs
@@ -1,0 +1,35 @@
+//! Manual workarounds for ostree bugs
+
+use std::io::Read;
+use std::ptr;
+
+use ostree;
+use ostree::prelude::{Cast, InputStreamExtManual};
+use ostree::{gio, glib};
+
+#[allow(unsafe_code)]
+
+/// Equivalent of `g_file_read()` for ostree::RepoFile to work around an ostree bug.
+pub fn repo_file_read(f: &ostree::RepoFile) -> Result<gio::InputStream, glib::Error> {
+    use glib::translate::*;
+    let stream = unsafe {
+        let f = f.upcast_ref::<gio::File>();
+        let mut error = ptr::null_mut();
+        let stream = gio::ffi::g_file_read(f.to_glib_none().0, ptr::null_mut(), &mut error);
+        if !error.is_null() {
+            return Err(from_glib_full(error));
+        }
+        let stream = stream as *mut gio::ffi::GInputStream;
+        from_glib_full(stream)
+    };
+
+    Ok(stream)
+}
+
+/// Read a repo file to a string.
+pub fn repo_file_read_to_string(f: &ostree::RepoFile) -> anyhow::Result<String> {
+    let mut r = String::new();
+    let mut s = repo_file_read(f)?.into_read();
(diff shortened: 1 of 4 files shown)


cgwalters:
Draft since this passes unit test CI, but I want to test it out for real on a host system and more carefully check that e.g. we're not doing anything stupid like physical file copies during the checkout/commit phase.

cgwalters:
OK yep, been playing with this some more and it looks like it's working quite well!

lib/src/ostree_manual.rs
```diff
@@ -0,0 +1,35 @@
...
+pub fn repo_file_read(f: &ostree::RepoFile) -> Result<gio::InputStream, glib::Error> {
+    use glib::translate::*;
+    let stream = unsafe {
+        let f = f.upcast_ref::<gio::File>();
+        let mut error = ptr::null_mut();
+        let stream = gio::ffi::g_file_read(f.to_glib_none().0, ptr::null_mut(), &mut error);
+        if !error.is_null() {
+            return Err(from_glib_full(error));
+        }
+        let stream = stream as *mut gio::ffi::GInputStream;
```
lucab, line 21: It would be good to make sure that `stream` is non-null before doing this.

lib/src/ostree_manual.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::io::Read;
+use std::ptr;
+
+use ostree;
+use ostree::prelude::{Cast, InputStreamExtManual};
+use ostree::{gio, glib};
+
+#[allow(unsafe_code)]
+
+/// Equivalent of `g_file_read()` for ostree::RepoFile to work around an ostree bug.
```
lucab, line 10: Let's explicitly link https://github.com/ostreedev/ostree/issues/2703 here maybe?

lucab (approved):

lib/src/ostree_manual.rs
cgwalters: We're just casting the pointer, not dereferencing here.  The dereference happens in `from_glib_full`, which already has a null check: https://docs.rs/glib/0.14.4/src/glib/shared.rs.html#422

cgwalters:
Also dropped the manual `g_file_query_info()` wrapper - I somehow confused myself into thinking that was necessary, but it isn't AFAICS.

PR #361, opened 2022-08-26
bootabletree: Add an API to find kernel in fs checkout too

cgwalters:
See https://github.com/coreos/rpm-ostree/pull/3966#discussion_r955281628

I want to switch over rpm-ostree to use this API instead of
reimplementing it.

This also adds test coverage for the existing GFile based API.


jmarrero (approved):
lgtm

PR #362, opened 2022-09-01
deploy: Add a `--no-imgref` option

cgwalters:
In https://github.com/coreos/coreos-assembler/pull/2523 we
taught coreos-assembler how to generate disk images with a
"pre-pulled" container image.  This means that the *first* OS update
will use shared layers.

However...right now running e.g. `rpm-ostree rebase quay.io/newimage`
won't necessarily prune the previous image.  (This may be considered
a bug)

But in practice, particularly for RHEL CoreOS we may not want to have
a default image reference - we don't (necessarily) want typing
`rpm-ostree upgrade` to do something.

With this, we can effectively pre-pull just the layers but not
the final image.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -802,6 +802,7 @@ where
                         kargs: kargs.as_deref(),
                         target_imgref: target_imgref.as_ref(),
                         proxy_cfg: Some(proxyopts.into()),
+                        ..Default::default()
                     };
                     let state = crate::container::deploy::deploy(
                         sysroot,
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

lucab:
It would be good to refresh the `quay.io/cgwalters/fcos-chunked` image and then re-trigger CI before merging:
```
+ ostree-ext-cli container image deploy --sysroot /run/host --stateroot testos --imgref ostree-unverified-registry:quay.io/cgwalters/fcos-chunked:v1
error: Performing deployment: remote error: reading manifest v1 in quay.io/cgwalters/fcos-chunked: unknown: Tag v1 was deleted or has expired. To pull, revive via time machine
```

cgwalters:
> It would be good to refresh the quay.io/cgwalters/fcos-chunked image and then re-trigger CI before merging:

The CI was already fixed in https://github.com/ostreedev/ostree-rs-ext/pull/363 - this just needed a rebase :surfer:

PR #364, opened 2022-09-09
container/commit: Actually return an error code

cgwalters:
I ended up rewriting a lot of this so we could more properly unit test it.  But the problem here boils down to `move ||` having *moved* the error counter into the closure, but since it's copy it made a new copy.

### DIFF
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -5,23 +5,25 @@
 use crate::container_utils::require_ostree_container;
 use anyhow::Context;
 use anyhow::Result;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
 use std::fs;
-use std::path::Path;
 use tokio::task;
 
 /// Check if there are any files that are not directories and error out if
 /// we find any, /var should not contain any files to commit in a container
 /// as it is where we expect user data to reside.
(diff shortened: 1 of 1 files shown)


cgwalters:
I actually did this minimal fix first:

```
diff --git a/lib/src/commit.rs b/lib/src/commit.rs
index 3f30842..51cd977 100644
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -39,15 +39,15 @@ pub(crate) async fn container_commit() -> Result<()> {
     println!("Checking /var for files");
     let var_path = Path::new("/var");
 
-    let mut error_count = 0;
-
     task::spawn_blocking(move || -> Result<()> {
-        validate_directories_only(var_path, &mut error_count)
+        let mut error_count = 0;
+        validate_directories_only(var_path, &mut error_count)?;
+        if error_count != 0 {
+            anyhow::bail!("Found content in /var");
+        }
+        Ok(())
     })
     .await??;
 
-    if error_count != 0 {
-        anyhow::bail!("Found content in /var");
-    }
     Ok(())
 }
```

but the desire to unit test this drove a larger rework.

jmarrero (approved):
lgtm

PR #366, opened 2022-09-13
container: Add an API to warn about non-ostree compatible content

cgwalters:
And use it consistently in both pull and deploy paths.

I also plan to make this very visible in rpm-ostree.

Motivated by a private chat discussion with someone not realizing that files put in `/var` in a container build would just be discarded.


jmarrero (approved):
lgtm

PR #367, opened 2022-09-13
container/commit: Autoclean directories, also handle `/run`

cgwalters:
The original command here was just scoped to `/var`, but we also don't want content in `/run`.  Extend the tooling to handle that.

Also, let's be a bit nicer here and auto-clean empty directories.

I was testing out the
https://github.com/coreos/coreos-layering-examples/blob/main/tailscale/Dockerfile example and today we have this:

```
drwxr-xr-x root/root         0 2022-09-13 18:53 run/
drwxr-xr-x root/root         0 2022-09-13 18:51 run/rpm-ostree/
drwxr-xr-x root/root         0 2022-09-13 18:53 run/rpm-ostree/lock/
drwxr-xr-x root/root         0 2022-09-13 18:51 run/systemd/
drwxr-xr-x root/root         0 2022-09-13 18:51 run/systemd/resolve/
-rwx------ root/root         0 2022-09-13 18:51 run/systemd/resolve/stub-resolv.conf
...
drwxr-xr-x root/root         0 2022-09-13 18:53 var/
drwxr-xr-x root/root         0 2022-09-13 18:53 var/cache/
drwx------ root/root         0 2022-09-13 18:53 var/cache/ldconfig/
-rw------- root/root     22000 2022-09-13 18:53 var/cache/ldconfig/aux-cache
drwxr-xr-x root/root         0 2022-09-08 23:10 var/cache/tailscale/
drwxr-xr-x root/root         0 2022-09-13 18:53 var/tmp/
```

In this set, we can auto-clean the leftover locking directories rpm-ostree (erroneously) leaves in `/run`.

I am tempted to auto-cleanup the `stub-resolv.conf` thing and the `ldconfig/aux-cache`.


cgwalters:
This pairs with https://github.com/ostreedev/ostree-rs-ext/pull/366

lucab:
I do like the empty directories /var autocleaning, it's a clear UX improvement. It assumes there is something re-creating those on boot (e.g. tmpfiles.d fragment or the software itself), but I think that's fine. And we can always implement auto-translation to tmpfiles.d here in the future if we want.

I'm less convinced about inspecting `/run`. IMHO it makes sense to leave that one alone (i.e. ignore any content landing there), basically treating it similarly to `/tmp` as an ephemeral hierarchy.

cgwalters:
> I'm less convinced about inspecting /run. IMHO it makes sense to leave that one alone (i.e. ignore any content landing there), basically treating it similarly to /tmp as an ephemeral hierarchy.

Good feedback!  Fixed.  For example, we no longer warn about anything in `/var/tmp` but just silently clean it, which definitely seems friendlier to me.

cgwalters:
Tangentially related, cc https://github.com/containers/buildah/issues/4242 for a podman bug we may also want to work around in this code.

lucab (approved):
Code LGTM, but CI needs some tweaking.

cgwalters:
Got the new CI green :tada:

PR #370, opened 2022-09-16
container/deploy: Add support for retaining

cgwalters:
This is needed for https://github.com/coreos/rpm-ostree/issues/4018 where we want to run the new ostree code from a container image.

Without this, ostree will think we're not booted into a deployment, and when we write a new one it will just happily delete the running root filesystem...


cgwalters:
https://github.com/ostreedev/ostree-rs-ext/pull/371 is the real fix here.  This makes sense on its own but let's merge that first.

cgwalters:
I don't think we really need this one.

PR #372, opened 2022-09-19
Release 0.8.6

cgwalters:
Mainly to get the deploy changes out so we can get a firstboot path in OCP that doesn't involve double reboot.


jmarrero (approved):
lgtm

Issue #154, opened 2021-11-09
oci export should support tags


cgwalters:
Currently e.g.
`rpm-ostree ex-container export --repo=/ostree/repo 953707641ddda1962b54b5e4b10ffa7d13abeede28452762762b4307f46b7aee oci:/var/tmp/fcos:latest`
creates `/var/tmp/fcos:latest` as a literal directory.

For now we should parse tags on our own, but the long term clear solution I think is for `skopeo experimental-image-proxy` to support writing too.

PR #374, opened 2022-09-20
container: Support writing tags to `oci:` directories

cgwalters:
I want to teach coreos-assembler to maintain an oci directory with version tags for builds instead of using an ostree repository.

It now works to do e.g.:

```
$ ostree-ext-cli container encapsulate --repo=tmp/repo 36.20220920.dev.0 oci:tmp/builds:36.20220920.dev.0
$ ostree-ext-cli container encapsulate --repo=tmp/repo 36.20220920.dev.1 oci:tmp/builds:36.20220920.dev.1
```

But in an OCI/container native build flow, we'd write the OCI builds not as `.ociarchive` but in an oci directory `builds-oci` or so.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/154

### DIFF
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -179,12 +179,14 @@ fn build_oci(
     repo: &ostree::Repo,
     rev: &str,
     ocidir_path: &Path,
+    tag: Option<&str>,
     config: &Config,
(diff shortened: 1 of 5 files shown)


jmarrero (approved):
lgtm

PR #377, opened 2022-09-21
 commit: Ignore files and directories not on the root mount

cgwalters:
ci: Also check via podman

It turns out that podman adds default mounts in `/run` which
means there's a behavior difference here.

But we want to cross check with both on general principle.  I was
just using `docker` because I thought podman might not be installed
on the stock GH action runner, but it apparently is.

---

commit: Ignore files and directories not on the root mount

podman injects e.g. `/run/.containerenv` as a tmpfs mount.  We should
not attempt to remove this - it will be gone when podman goes to
write the tar stream itself.

Also in the general case, traversing and recursively removing
is potentially really dangerous if e.g. someone mounted an external
drive during a container build process at say `/var/srv`.  Hopefully
most of those cases would be read-only, but still.  Ultimately
the point is that we only want to remove files that will become
part of the serialized tar stream.

Pass the device ID of `/` down into our traversal and ignore
files/directories which don't match.

---

### DIFF
--- a/ci/container-build-integration.sh
+++ b/ci/container-build-integration.sh
@@ -13,8 +13,9 @@ sed -ie 's,ostree container commit,ostree-ext-cli container commit,' Dockerfile
 sed -ie 's,^\(FROM .*\),\1\nADD ostree-ext-cli /usr/bin,' Dockerfile
 git diff
 
-docker build -t localhost/fcos-tailscale .
-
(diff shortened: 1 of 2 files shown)


cgwalters:
>  error: Device or resource busy (os error 16)

OK cool, new `podman` test successfully reproduced the problem; added the fix.

jmarrero (approved):
/lgtm

PR #378, opened 2022-09-21
Add compat feature and bump semver to 0.9

cgwalters:
Add a `compat` feature

This is the first step towards disabling our support for old
tar formats and the legacy container formats.

When the `compat` feature is off, everything defaults to v1
format, and we reject parsing the old v0 container images in
particular.

---

Bump semver to 0.9

Because we have an off-by-default `compat` feature.

---


cgwalters:
Hmm..  Perhaps it's simpler to start by adding a "loudly warn and sleep if we find old images" API...

cgwalters:
OK, rebased :surfer: and fixed the tests.

jmarrero (approved):
/lgtm

PR #381, opened 2022-09-29
cargo: bump to clap v4

lucab:
RFC: clap v4 comes with several additional deprecations, which I haven't yet looked at. We can either do it here, or leave it for later.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -12,7 +12,7 @@ rust-version = "1.64.0"
 [dependencies]
 anyhow = "1.0"
 ostree-ext = { path = "../lib" }
-clap = "3.2"
+clap = "4.2"
 libc = "0.2.92"
 tokio = { version = "1", features = ["macros"] }
 log = "0.4.0"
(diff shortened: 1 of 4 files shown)


lucab:
```
error: failed to select a version for the requirement `clap = "^4.0"`
candidate versions found which didn't match: 3.2.22, 3.2.21, 3.2.20, ...
location searched: crates.io index
```

This is a funny new cargo error that I've never seen before. It seems to only happen on our old/pinned toolchains (1.58.1).

cgwalters:
I think it's because cargo is skipping the dependency since our MSRV doesn't match the `rust=` in the crate.

lucab:
I think this is because of this new syntax for Cargo features that `clap` is using: https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features
Our pinned toolchain is older than 1.60 so it doesn't understand it.

cgwalters:
Hmm, I am not sure how to get clap_mangen to work in the new flow.  Seems we can't call `clone()` on a `Command` anymore - clap panics about a duplicate name.

cgwalters:
I'm not seeing any new deprecations now...it seems likely that clap upstream now as of 4.2 backed off that?

cgwalters (approved):

Issue #380, opened 2022-09-28
container detection fails inside `docker buildx`


cgwalters:
I was playing around with GH actions for container building, it uses `docker buildx` by default and the environment there appears to lack `/.dockerenv`.

Hmmm.  It's a bit tempting to force on setting the `container` environment variable in ostree native containers.

vrutkovs:
I've hit this in github actions in https://github.com/vrutkovs/custom-okd-os and the fix was to add
`ENV container oci` in the dockerfile

PR #382, opened 2022-09-30
container: Loosen container detection to work with `docker buildx`

cgwalters:
Apparently, `docker buildx build` creates containers without `/.dockerenv`.  I debated forcing on a `container=oci` environment variable for our containers, but upon some reflection I think it's about equally as clean to loosen our container detection logic.

The presence of a bare-split-xattrs ostree repo is a powerful signal. Let's use that as the signal, unless overridden by strong contrary evidence:

- We're running under systemd
- We are apparently on a booted ostree systemd

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/380

### DIFF
--- a/ci/container-build-integration.sh
+++ b/ci/container-build-integration.sh
@@ -6,16 +6,27 @@ image=quay.io/coreos-assembler/fcos:stable
 example=coreos-layering-examples/tailscale
 set -x
 
-mv ostree-ext-cli ${example}
-cd ${example}
 chmod a+x ostree-ext-cli
+workdir=${PWD}
+cd ${example}
+cp ${workdir}/ostree-ext-cli .
 sed -ie 's,ostree container commit,ostree-ext-cli container commit,' Dockerfile
-sed -ie 's,^\(FROM .*\),\1\nADD ostree-ext-cli /usr/bin,' Dockerfile
+sed -ie 's,^\(FROM .*\),\1\nADD ostree-ext-cli /usr/bin/,' Dockerfile
 git diff
 
 for runtime in podman docker; do
     $runtime build -t localhost/fcos-tailscale .
     $runtime run --rm localhost/fcos-tailscale rpm -q tailscale
 done
+
+cd $(mktemp -d)
+cp ${workdir}/ostree-ext-cli .
+cat > Dockerfile << EOF
+FROM $image
+ADD ostree-ext-cli /usr/bin/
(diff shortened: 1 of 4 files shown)


cgwalters:
@lucab asked whether we could rely on the `container` environment variable, but no, docker has always resisted setting that (for some reason).

I pushed a test change https://github.com/cgwalters/ostree-rs-ext/commit/baa3c9b2be0e76b1477570584d287437bb273eb3
and you can [see the result over here](https://github.com/cgwalters/ostree-rs-ext/actions/runs/3159303276/jobs/5142462022):

```
 #7 [3/3] RUN env && set -x; test $(ostree-ext-cli internal-only-for-testing detect-env) = ostree-container
#7 0.260 ***
#7 0.260 HOME=/root
#7 0.260 SHLVL=1
#7 0.260 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
#7 0.260 _=/usr/bin/env
```

So as I noted in the PR *we* could force on the `container=` environment variable for our containers, and I started to do so but I think this is in the end better actually.

jlebon:
Another reference in this field I've looked at in the past is systemd: https://github.com/systemd/systemd/blob/e879434df550c6dcfc02f23e613f4dda7f741089/src/basic/virt.c#L645. It needs this to implement `ConditionVirtualization`. What does `systemd-detect-virt` say?

cgwalters:
To be clear, I'm not opposed to defaulting to injecting a `container=` environment variable.  But it'd require respinning all of our existing containers, and it's something which hasn't been adopted broadly by the whole ecosystem.  We can do it as a next step at a later point.

Again the rationale of this PR is we have a pretty unambiguous thing on the *filesystem* that tells us we're in an ostree-container, which is what we care about.

cgwalters:
One thing I wanted to say here but forgot is an important motivation behind this PR is that the default Github action for building containers uses `docker buildx`, so with this we start working in that environment which is quite important to me.

jlebon (approved):
Given that this checks for the special repo mode in the special repo location, using it as a signal seems reasonable to me. I didn't review the test changes in depth, but LGTM!

cgwalters:
FWIW the answer is:

```
#8 [4/5] RUN systemd-detect-virt
#8 0.288 container-other
```

Which I think is coming from https://github.com/systemd/systemd/blob/e879434df550c6dcfc02f23e613f4dda7f741089/src/basic/virt.c#L553

We could consider actually executing `systemd-detect-virt` (if available) as a followup too.

PR #385, opened 2022-10-07
Add warning infrastructure when we find v0 images

cgwalters:
Let's make this much more visible by exposing the ability to do warn-and-sleep.


jmarrero (approved):
lgtm

PR #387, opened 2022-10-18
Automatically log some container metadata to systemd journal (if avail)

cgwalters:
I was logged into one of my OpenShift machines live and watching `journalctl -b -u rpm-ostreed -f` during an upgrade, and was surprised not to see any useful output.

This is because of course rpm-ostreed logs everything to the client, in this case the MCD.

I think we should also duplicate these messages directly in the journal.  libostree does something similar when fetching via HTTP.

### DIFF
--- a/lib/src/logging.rs
+++ b/lib/src/logging.rs
@@ -1,0 +1,37 @@
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// Set to true if we failed to write to the journal once
+static EMITTED_JOURNAL_ERROR: AtomicBool = AtomicBool::new(false);
+
+/// Wrapper for systemd structured logging which only emits a message
+/// if we're targeting the system repository, and it's booted.
+pub(crate) fn system_repo_journal_send<K, V>(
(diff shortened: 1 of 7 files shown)


jmarrero (approved):
lgtm

lib/src/logging.rs
```diff
@@ -0,0 +1,37 @@
...
+    msg: &str,
+    vars: impl Iterator<Item = (K, V)>,
+) where
+    K: AsRef<str>,
+    V: AsRef<str>,
+{
+    if !repo.is_system() {
+        return;
+    }
+    if let Err(e) = libsystemd::logging::journal_send(priority, msg, vars) {
```
lucab, line 21: I think you may be interested in https://docs.rs/libsystemd/0.5.0/libsystemd/logging/fn.connected_to_journal.html here. Although I'm not sure if RHEL 8 supports that.

cgwalters: In this case I think it's actually a desired feature that we log to the journal even if directly executed from a root shell (i.e. not in a systemd unit).

PR #393, opened 2022-11-03
cli: Add `--no-imgref` to deploy option

cgwalters:
Previously I added it to the API, intending to use it for FCOS, but hadn't actually tested that at the time.

I'm still not entirely sure we will, but let's add it to the CLI anyways.


jmarrero (approved):
lgtm

PR #394, opened 2022-11-03
logging: Don't write to journal if not running systemd

cgwalters:
I noticed an error message when we're doing the deployment in supermin in coreos-assembler.


cgwalters:
(Of course...we actually I think should be running systemd there, the supermin thing is just weird and broken...but, anyways)

cgwalters:
OK, got the tests passing on this one.

lucab (approved):

Issue #153, opened 2021-11-09
Use diffids - fix pulling from `containers-storage:`


cgwalters:
Today we're using https://github.com/containers/containers-image-proxy-rs which is a Rust binding to https://github.com/containers/skopeo/blob/main/cmd/skopeo/proxy.go which is an IPC interface to containers/image.

In our usage of this, we fetch by blob right now.  Operating system upgrades for example work fine fetching from a remote registry (e.g. `quay.io/fedora/fedora-coreos:stable`) - we stream and uncompress the blobs, writing to ostree.

However, it doesn't work to fetch by blob ID (compressed tar) from `containers-storage` because it only has the uncompressed version.  We need to fetch via diffid.

We need https://github.com/containers/skopeo/pull/1495 to fetch the config blob, then the pull code should use it.

mtrmac:
Alternatively the proxy could expose LayerInfosForCopy, to be closer to what the c/image/copy code does and assumes.

(Admittedly LayerInfosForCopy is a wart that shouldn’t exist in the first place… between exposing that and hard-coding `containers-storage:` in clients, I find it hard to decide which one I dislike more :) )

cgwalters:
Just noting that https://github.com/RishabhSaini is planning to look at this

nalind:
I'd second the suggestion to expose and use `LayerInfosForCopy()`, and then add `LayerInfos()` to it as well.

A v2s1 image isn't going to have any DiffID values in the OCI-format config blob that is synthesized for it (observe: `skopeo inspect --config docker://registry.k8s.io/pause:3.0`), so reading the config and assuming that the DiffID values it contains can be used as suitable input values for `GetBlob()` is not always going to work.

`LayerInfos()` and `LayerInfosForCopy()` can supply the set of values that `GetBlob()` is prepared to be asked for.

cgwalters:
- Add a proxy API to serialize the result of `LayerInfosForCopy` (if returns nothing, call `GetLayerInfos`)
- We can probably directly just call `json.Marshal` on this and have it work (but it'd be better to add json tags to the struct most likely)
- We can call `GetBlob` on these digests, filtering out the blobs (compressed) that we already have.
- We may need to store the uncompressed blob ID in the ostree commit that stores the blob to make this more efficient

PR #396, opened 2022-11-03
Fix pulling from containers-storage

RishabhSaini:
fixes: #153
Uses
containers-image-proxy-rs: https://github.com/containers/containers-image-proxy-rs/pull/37
skopeo: https://github.com/containers/skopeo/pull/1802

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -583,7 +583,7 @@ impl ImageImporter {
                 None
             }
         };
-
+        let des_layers = self.proxy.get_layer_info(&self.proxy_img).await?;
         for layer in import.ostree_layers.iter_mut() {
             if layer.commit.is_some() {
                 continue;
@@ -598,6 +598,8 @@ impl ImageImporter {
                 &import.manifest,
                 &layer.layer,
                 self.layer_byte_progress.as_ref(),
+                des_layers.as_ref(),
+                self.imgref.imgref.transport,
             )
             .await?;
             let repo = self.repo.clone();
@@ -641,6 +643,8 @@ impl ImageImporter {
                 &import.manifest,
                 &import.ostree_commit_layer.layer,
                 self.layer_byte_progress.as_ref(),
+                des_layers.as_ref(),
+                self.imgref.imgref.transport,
             )
             .await?;
             let repo = self.repo.clone();
@@ -712,6 +716,7 @@ impl ImageImporter {
         // First download all layers for the base image (if necessary) - we need the SELinux policy
         // there to label all following layers.
         self.unencapsulate_base(&mut import, true).await?;
+        let des_layers = self.proxy.get_layer_info(&self.proxy_img).await?;
         let mut proxy = self.proxy;
         let target_imgref = self.target_imgref.as_ref().unwrap_or(&self.imgref);
         let base_commit = import.ostree_commit_layer.commit.clone().unwrap();
@@ -735,6 +740,8 @@ impl ImageImporter {
                     &import.manifest,
                     &layer.layer,
                     self.layer_byte_progress.as_ref(),
+                    des_layers.as_ref(),
+                    self.imgref.imgref.transport,
                 )
                 .await?;
                 // An important aspect of this is that we SELinux label the derived layers using
--- a/lib/src/container/unencapsulate.rs
+++ b/lib/src/container/unencapsulate.rs
@@ -205,6 +205,8 @@ pub(crate) async fn fetch_layer_decompress<'a>(
     manifest: &oci_image::ImageManifest,
     layer: &'a oci_image::Descriptor,
     progress: Option<&'a Sender<Option<store::LayerProgress>>>,
+    layer_info: Option<&Vec<containers_image_proxy::ConvertedLayerInfo>>,
+    transport_src: Transport,
 ) -> Result<(
     Box<dyn AsyncBufRead + Send + Unpin>,
     impl Future<Output = Result<()>> + 'a,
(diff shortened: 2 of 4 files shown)


lib/src/container/store.rs
```diff
@@ -546,10 +594,13 @@ impl ImageImporter {
             }
             let (blob, driver) = fetch_layer_decompress(
                 &mut self.proxy,
-                &self.proxy_img,
+                self.proxy_img.clone(),
```
cgwalters, line 597: We don't want to have multiple proxy processes.

lib/src/container/store.rs
```diff
@@ -413,6 +449,18 @@ impl ImageImporter {
         })
     }
 
+    ///Given a target diffid, return its corresponding blob
+    //Not fetch all the blobs
+    //How do we use diffid here?
+    pub async fn blob_from_diffid(
+        self,
```
cgwalters, line 456: I think the problem is here, you want this to be `&self` to borrow instead of move.

lib/tests/it/main.rs
```diff
@@ -961,6 +961,26 @@ async fn test_container_import_export_v1() {
...
 }
 
+
+///Test layers
+#[tokio::test]
+async fn test_layers() -> Result<()> {
+    let fixture = Fixture::new_v1()?;
+    let srcpath = fixture.path.join("src/repo");
+    let srcrepo_parsed = ostree_ext::cli::parse_repo(&srcpath).unwrap();
+    let ir = ostree_ext::cli::parse_imgref("ostree-remote-image:containers-storage:quay.io/fedora/fedora-coreos:testing-devel")
```
cgwalters, line 971: This string won't work; I think you mean e.g. `ostree-unverified-image:containers-storage:quay.io/fedora/fedora-coreos:testing-devel`.  But even then, I don't think we can do this in a unit test.  A general idea is that unit tests shouldn't change any global system state, and pulling a container image into e.g. `~/.local/share/containers` is global system state.

We could make that type of test opt in?  Alternatively we have e.g. `integrationtest.rs` - could make sense there.

lib/src/container/unencapsulate.rs
```diff
@@ -202,17 +201,34 @@ pub(crate) async fn fetch_layer_decompress<'a>(
...
+    transport_src: Transport
 ) -> Result<(
     Box<dyn AsyncBufRead + Send + Unpin>,
     impl Future<Output = Result<()>> + 'a,
 )> {
     use futures_util::future::Either;
     tracing::debug!("fetching {}", layer.digest());
     let layer_index = manifest.layers().iter().position(|x| x == layer).unwrap();
+    let (blob, driver, size);
+    let mediaType : &oci_image::MediaType;
```
cgwalters, line 214: I think you need to `cargo fmt`.

lib/src/container/store.rs
```diff
@@ -712,6 +716,7 @@ impl ImageImporter {
         // First download all layers for the base image (if necessary) - we need the SELinux policy
         // there to label all following layers.
         self.unencapsulate_base(&mut import, true).await?;
+        let des_layers = self.proxy.get_layer_info_json(&self.proxy_img).await?;
```
cgwalters, line 719: This is going to introduce a hard requirement on a new skopeo; I'm not sure we can do that yet.  I see two options:

- Add a cargo feature - for precedence, look at https://github.com/containers/containers-image-proxy-rs/pull/36 and cdd20ecf1f0097c485c318d6b757b6ea074db845 - I can help with this.  
- Call this API only if we're fetching from `containers-storage:` (If we're not going to add a static feature, then we should have the proxy return e.g. an `Option<LayersInfo>` so we can *dynamically* check for support)

lib/src/integrationtest.rs
```diff
@@ -101,6 +101,27 @@ where
     Ok(())
 }
 
+async fn test_pull_containers_storage() -> Result<()> {
+    let fixture = crate::fixture::Fixture::new_v1()?;
+    let srcpath = fixture.path.join("src/repo");
+    let srcrepo_parsed = crate::cli::parse_repo(&srcpath).unwrap();
+    let ir = crate::cli::parse_imgref(
+        "ostree-unverified-image:containers-storage:quay.io/fedora/fedora-coreos:testing-devel",
```
cgwalters, line 109: This assumes this image has already been fetched, but we'd need to do that explicitly in the test flow right?

I'm also realizing that we probably want to split out a different kind of integration test...this one as you've noted assumes it's already running in a container image.

It may be simpler to do one in shell script.  Hmm, we already have `ci/priv-integration.sh` and ISTM in theory we could potentially just extend that?  It might be though that we'd need to bind mount some other things from the host to make it work (e.g. `/var/lib/containers`).  Nested container stuff is messy...I guess we could also try to have a CI flow that builds natively for Ubuntu, but that brings its own mess.

The other alternative is to use virtualization for this, i.e. we add Prow or CoreOSCI to this repo.  Which is probably overdue.

cgwalters:
I think we want to rebase this on https://github.com/ostreedev/ostree-rs-ext/pull/397 and also bump the containers-image-proxy version.

cgwalters:
I think https://crates.io/crates/containers-image-proxy/0.5.2 has your change, but it also adds a dependency on cap-std 1.0 so it'd be good to dedup on #397 so that we only pull in one version of that.

But alternatively just adding a bump of the imageproxy in this PR should work.

cgwalters:
The IMA failure was a flake (Fedora rpm repo infrastructure), I just restarted the job via the Actions UI.

lib/src/container/unencapsulate.rs
```diff
@@ -205,17 +205,38 @@ pub(crate) async fn fetch_layer_decompress<'a>(
...
     impl Future<Output = Result<()>> + 'a,
 )> {
     use futures_util::future::Either;
     tracing::debug!("fetching {}", layer.digest());
     let layer_index = manifest.layers().iter().position(|x| x == layer).unwrap();
+    let (blob, driver, size);
+    let media_type: &oci_image::MediaType;
+    match transport_src {
+        Transport::ContainerStorage => {
+            let n_layers = layer_info.unwrap().len();
```
cgwalters, line 221: Instead of `.unwrap()` here, let's do:

```suggestion
           let layer_info = layer_info.ok_or_else(|| anyhow!("skopeo is too old to pull from containers-storage"))?;
           let n_layers = layer_info.len();
```
etc.  (You'll also want to drop the `unwrap()` from the line below)

cgwalters: Then what you can do for now is adjust the integration test to look for the string "skopeo is too old".

Once the skopeo PR lands and is built, we can change the integration test again to finally verify things work!

RishabhSaini:
Updated

cgwalters (approved):
Nice!

PR #398, opened 2022-11-04
Two minor patches

cgwalters:
cli: Don't publish

We aren't productizing this CLI entrypoint.

---

Don't expose `ostree_manual` API at all in non-test mode

This avoids dead code warnings for external crate users.

---


jmarrero (approved):
lgtm

PR #399, opened 2022-11-04
Initial support for generating manpages

cgwalters:
It's about time we had online (and shipped in packages) docs for our tools.

Some code inspired/taken from https://github.com/coreos/coreos-installer/blob/main/src/cmdline/doc.rs


jmarrero (approved):
lgtm

PR #400, opened 2022-11-08
tar/export: Don't create hardlinks to zero sized files

cgwalters:
This is an echo of https://github.com/ostreedev/ostree/pull/2198/commits/558720e7aa1870cbbdb4a0dc22a3968d116daec3

If we hardlink zero sized files, then any modification to them will result in possibly *many* hardlinks needing to be reproduced when serializing to tar, which is definitely suboptimal.

But further, it provokes a bug in our import path; when processing derived layers, we need to handle *both* cases of having the file in `/sysroot` be the source as well as the destination of the hardlink in the tar stream.


cgwalters:
Draft since this needs some tests, but I did verify it fixes https://github.com/coreos/rpm-ostree/issues/3827

cgwalters:
> Draft since this needs some tests

Added a basic test.

jmarrero (approved):
lgtm

Issue #405, opened 2022-11-10
handle modified hardlinks to `/sysroot`


cgwalters:
Today this:

```
FROM quay.io/fedora/fedora-coreos:stable
RUN touch /usr/bin/bash
```

When built using e.g. the `redhat-actions/buildah-build@v2` Github action, seems to result in a tar stream that looks like this:

```
drwxr-xr-x root/root         0 1969-12-31 19:00 sysroot/
drwxr-xr-x root/root         0 1969-12-31 19:00 sysroot/ostree/
drwxr-xr-x root/root         0 1969-12-31 19:00 sysroot/ostree/repo/
drwxr-xr-x root/root         0 1969-12-31 19:00 sysroot/ostree/repo/objects/
drwxr-xr-x root/root         0 1969-12-31 19:00 sysroot/ostree/repo/objects/60/
-rwxr-xr-x root/root   1390000 2022-11-08 18:26 sysroot/ostree/repo/objects/60/feb13e826d2f9b62490ab24cea0f4a2d09615fb57027e55f713c18c59f4796.file
drwxr-xr-x root/root         0 2022-11-08 18:26 usr/
drwxr-xr-x root/root         0 2022-11-08 18:26 usr/bin/
hrwxr-xr-x root/root         0 2022-11-08 18:26 usr/bin/bash link to sysroot/ostree/repo/objects/60/feb13e826d2f9b62490ab24cea0f4a2d09615fb57027e55f713c18c59f4796.file
```

But that causes problems because when importing layered builds, we discard content in `/sysroot` entirely.

I think this has something to do with the use of the overlayfs driver versus vfs.  I believe with the kernel overlayfs, modifying a hardlinked file will result in a copyup of *just that single file*; it explicitly will not mutate all targets of that link.  From [the docs](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt):

> 2) "inode index"
> Enabled with the mount option or module option "index=on" or with the
kernel config option CONFIG_OVERLAY_FS_INDEX=y.
> If this feature is disabled and a file with multiple hard links is copied
up, then this will "break" the link.  Changes will not be propagated to
other names referring to the same inode.

I can think of two fixes:

- During `ostree container commit`, we can actually nuke everything in `/sysroot` - or at least just files whose timestamp is newer than the epoch.  This would result in whiteout files in the serialized stream, which the import path will just ignore.
[...]

PR #408, opened 2022-11-11
tar/import: Handle hardlinked changes into /sysroot

cgwalters:
Builds on https://github.com/ostreedev/ostree-rs-ext/pull/406

---

tar/import: Handle hardlinked changes into /sysroot

Some container build processes run without overlayfs inode indexing
on - in this common scenario, overlayfs is not quite POSIX compliant
because it will break the hardlink instead of modifying all versions.

We need to handle this case of having *all* names for a hardlinked
file being modified though too.  If the serialized tar stream
has the file in `/sysroot` be the canonical version, then because
we drop out that file here, we'll fail to import.

Fix this by significantly beefing up the tar filtering/reprocessing
logic:

 - When we see a *modified* file in `/sysroot` with a nonzero timestamp,
   cache its data into a lookaside temporary directory
 - If we then see a hardlink to that file path, make *that* file
   be the canonical version in e.g. `/usr`.
 - Any further hardlinks to `/sysroot` instead become hardlinks
   to the new canonical one.

(Arguably perhaps...we should actually not have used hardlinks
 in ostree containers at all, but injected this metadata in
 some other way.  But, the ship has sailed on that)

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/405

---

### DIFF
--- a/lib/src/integrationtest.rs
+++ b/lib/src/integrationtest.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use crate::{
     container::{ocidir, ExportLayout},
     container_utils::is_ostree_container,
+    ocidir::RawLayerWriter,
 };
 use anyhow::Result;
 use camino::Utf8Path;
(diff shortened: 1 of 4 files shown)


cgwalters:
We could probably just merge this directly and close https://github.com/ostreedev/ostree-rs-ext/pull/406 - but I wanted to demonstrate in 406 that the new test case was "successfully failing".

jmarrero (approved):
lgtm

PR #409, opened 2022-11-11
Release 0.9.1

cgwalters:
This has two major fixes:

- Better support for image pruning
- Handling of hardlinked files into `/sysroot` which can occur with some container build systems


jmarrero (approved):
lgtm

PR #410, opened 2022-11-14
container: Make single image removal idempotent

cgwalters:
Due to bugs or logic errors, it's possible that we might attempt to prune an image that's already been pruned.  In order to make things more robust, let's avoid making this a hard error.

Keep the semantics for the existing "prune multiple images" API, but add a new one that prunes a singular image, and does not error in the not-found case.  This can be used by higher level software.


lib/src/container/store.rs
```diff
@@ -1125,7 +1114,22 @@ pub fn image_filtered_content_warning(
...
+/// not present, this function will successfully perform no operation.
+///
+/// This function assumes no transaction is active on the repository.
+/// The underlying layers are *not* pruned; that requires a separate invocation
+/// of [`gc_image_layers`].
+#[context("Pruning {img}")]
+pub fn remove_image(repo: &ostree::Repo, img: &ImageReference) -> Result<bool> {
+    let ostree_ref = &ref_for_image(img)?;
+    let found = repo.resolve_rev(ostree_ref, true)?.is_some();
+    repo.set_ref_immediate(None, ostree_ref, None, gio::NONE_CANCELLABLE)?;
```
lucab, line 1127: Nit: just curious, is there a reason why you decided to call this also in the not found case? I know you mentioned this is idempotent, but I guess we could just skip an additional trip through libostree in that case.

lucab (approved):

lib/src/container/store.rs
cgwalters: Good point, adjusted!

PR #412, opened 2022-11-16
Add a "bootc" CLI

cgwalters:
This is a new CLI that is intended to be "container native", without the baggage of `ostree` (and `rpm-ostree`).

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2,12 +2,26 @@
 #![forbid(unused_must_use)]
 #![deny(unsafe_code)]
 
+use std::os::unix::prelude::OsStrExt;
+use std::path::Path;
+
 use anyhow::Result;
 
 async fn run() -> Result<()> {
     tracing_subscriber::fmt::init();
     tracing::trace!("starting");
-    ostree_ext::cli::run_from_iter(std::env::args_os()).await
+    let args = std::env::args_os().collect::<Vec<_>>();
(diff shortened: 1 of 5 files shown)


cgwalters:
This still wants some tests; split a prep patch into https://github.com/ostreedev/ostree-rs-ext/pull/413

cgwalters:
OK and now this one is rebased :surfer: - several of the helper bits invented here have been moved more properly into other places (e.g. the authfd bits, sysroot locking).  The main one that isn't yet is `async fn ensure_self_unshared_mount_namespace() -> Result<()> {` because that needs https://github.com/ostreedev/ostree/pull/2779 which needs a new ostree.  (I may investigate explicitly pulling that in in CI soon though).

cgwalters:
OK so...I'm thinking what may actually work out better here is if we move this to a cleanly separate git repository - say e.g. `github.com/containers/bootc`.  And we frame this as the *initial backend* happens to be ostree.  But perhaps we also support other things in the future.

cgwalters:
OK I've created https://github.com/cgwalters/bootc

cgwalters:
Closing this, the new plan is to keep this code living in a separate repo per above.

cgwalters:
@lucab as a current maintainer here, I do want to invite you to also be a maintainer on https://github.com/containers/bootc if you're interested!

PR #397, opened 2022-11-04
Bump semver to 0.10, ostree 0.17 (and glib 0.16) and cap-std 1.0

cgwalters:
This requires https://github.com/gtk-rs/gtk-rs-core/pull/820

PR #413, opened 2022-11-17
Add an isolation module, use it for pulls by default

cgwalters:
This effectively lowers into this project code from https://github.com/coreos/rpm-ostree/pull/3937/commits/d661e8f974f8d2550a865c2866476160e333ec72

We want to do this by default; we use the `nobody` user here.

### DIFF
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -26,6 +26,7 @@
 //! for this is [planned but not implemented](https://github.com/ostreedev/ostree-rs-ext/issues/12).
 
 use anyhow::anyhow;
+
 use std::borrow::Cow;
 use std::ops::Deref;
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


lib/src/container/mod.rs
```diff
@@ -239,8 +243,22 @@ impl std::fmt::Display for OstreeImageReference {
 
 /// Apply default configuration for container image pulls to an existing configuration.
 /// For example, if `authfile` is not set, and `auth_anonymous` is `false`, and a global configuration file exists, it will be used.
+///
+/// If the
```
lucab, line 247: Somebody stole the last part of the docstring here! :detective:

lib/src/isolation.rs
```diff
@@ -0,0 +1,44 @@
+use std::process::Command;
+
+use once_cell::sync::Lazy;
+
+pub(crate) const DEFAULT_UNPRIVILEGED_USER: &str = "nobody";
```
lucab, line 5: Let's directly use `65543` here maybe? That would avoid any messy entanglement with userdb lookups, and we already know we have a bit of crazy setup with altfiles and legacy `nobody` user.

cgwalters: Hmm.  Userdb lookups really have to work though.

lib/src/container/mod.rs
```diff
@@ -251,6 +269,30 @@ pub fn merge_default_container_proxy_opts(
...
+            if let Some(authfile) = config.authfile.take() {
+                let mut authfile = std::fs::File::open(&authfile).map(std::io::BufReader::new)?;
+                let tmp =
+                    &cap_std::fs::Dir::open_ambient_dir("/tmp", cap_std::ambient_authority())?;
+                let mut authfd =
+                    cap_tempfile::TempFile::new_anonymous(tmp).map(std::io::BufWriter::new)?;
+                std::io::copy(&mut authfile, &mut authfd)?;
+                let authfd = authfd.into_inner()?;
+                let authfd = std::sync::Arc::new(authfd.into_fd().into());
+                let target_fd = 5;
```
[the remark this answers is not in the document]
cgwalters: Well, nothing today...but OK, let's clean this up more by lowering the code one more level!

https://github.com/containers/containers-image-proxy-rs/pull/39

lib/src/isolation.rs
cgwalters: That said...well, we could try to cross the bridge of adding a requirement on a systemd service that uses `DynamicUser=ostree`...

cgwalters:
OK updated to depend on https://github.com/containers/containers-image-proxy-rs/pull/39

cgwalters:
OK, now that we have the new containers-image-proxy from https://github.com/ostreedev/ostree-rs-ext/pull/397 this should be good to go!

jmarrero (approved):
lgtm

PR #414, opened 2022-11-18
Ensure integration testing fails for pulling from containers-storage 

RishabhSaini:
Prep commit for #396 
Fails for: ostree-ext-cli container image pull $osrepo ostree-unverified-image:containers-storage:$image

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -4,6 +4,9 @@
 # whatever we want, however we can't actually *reboot* the host.
 set -euo pipefail
 
+# https://github.com/ostreedev/ostree-rs-ext/issues/417
+mkdir -p /var/tmp
+
 sysroot=/run/host
 # Current stable image fixture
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
When I try this locally *without* having the image pulled I get:

```
error: remote error: reference "[overlay@/var/lib/containers/storage+/run/containers/storage:overlay.mountopt=nodev,metacopy=on]quay.io/fedora/fedora-coreos:testing-devel" does not resolve to an image ID: identifier is not an image
```

So we need to e.g. `podman pull quay.io/fedora/fedora-coreos:testing-devel` first.

RishabhSaini:
I am currently running the checks in my forked repo to ensure it passes. I will update here when it passes.

RishabhSaini:
> I am currently running the checks in my forked repo to ensure it passes. I will update here when it passes.

Now it passes all tests.

PR #415, opened 2022-11-21
container: Add support for copying optionally-present keys

cgwalters:
This is to aid https://github.com/coreos/coreos-assembler/pull/3214 which is trying to inject the metadata key `fedora-coreos.stream` into the container image.  However, this value will only be present in Fedora derivatives, and not RHEL/CentOS.

Add support for copying a key only if present, instead of erroring if it's missing.


cgwalters:
Inbound fix for the IMA tests in https://github.com/ostreedev/ostree-rs-ext/pull/419

PR #416, opened 2022-11-22
Add a `sysroot` module with helper for `SysrootLock`

cgwalters:
A common use case is wanting to acquire a locked system root.  Add a helper function which handles this and returns a guard object that can be dereferenced to a sysroot.


jmarrero (approved):
lgtm

PR #418, opened 2022-11-22
ci: Switch to checkout@v3

cgwalters:
> Node.js 12 actions are deprecated. For more information see:
> https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


jmarrero (approved):
lgtm

Issue #417, opened 2022-11-22
ostree vs containers and `/var/tmp`


cgwalters:
By default in our generated containers, `/var` starts out empty.  This means `/var/tmp` is missing, which can understandably confuse some software.

I think I'd say we should at least synthesize `/var/tmp` in our filesystem.

PR #422, opened 2022-11-28
tar/export: Always generate `/var/tmp`

cgwalters:
This is a bit of a hack, but basically the ostree model of empty `/var` defers creation of basic directory structure to e.g. `systemd-tmpfiles`, but that isn't run in containers.

For now, let's keep this as the only special case.  Perhaps down the line we may actually need to effectively parse/execute systemd-tmpfiles inside the stream in the future, but hopefully we can avoid that.

Many things are going to use `/var/tmp` during image builds, but that's much less likely with e.g. `/var/log`.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/417


jmarrero (approved):
lgtm

PR #423, opened 2022-11-29
tests: Always use v1 for compat testing

cgwalters:
With a new enough rpm-ostree, this test will always fail if we're built in compat mode.  I'm hitting this locally, but it looks like the latest rpm-ostree only made it to testing-devel 2 days ago.


jmarrero (approved):
lgtm

PR #424, opened 2022-11-29
Drop `compat` feature

cgwalters:
Most recent rpm-ostree doesn't enable the compat feature and no one yelled, so let's pull off the band-aid here and drop the old format logic.


jmarrero (approved):
lgtm

PR #425, opened 2022-11-30
Two minor bits

cgwalters:
sysroot: Make underlying value `pub`

Since we expose it via `Deref` anyways.  Prep for usage in bootc.

---

impl Display for SignatureSource

This is the cleaner way to do things, and I want this in bootc.

---


jmarrero (approved):
lgtm

PR #426, opened 2022-12-01
store: Tweak status message for layers

cgwalters:
The formatting was confusing before because one could associate the number both before and after "layers".


jmarrero (approved):
lgtm

PR #431, opened 2022-12-05
store: Add some `tracing::` to GC paths

cgwalters:
To aid future debugging.  We should probably add some of this into the systemd journal always too, but that can be done in the future.


jmarrero (approved):
lgtm

PR #432, opened 2022-12-05
store: Add rpm-ostree base refs into GC set

cgwalters:
Closes: https://github.com/coreos/rpm-ostree/issues/4185

Basically, with package layering or client side commits, rpm-ostree generates synthetic refs under `rpmostree/base` which point to the base commits.  But this interacts badly with our image pruning logic.

I'm happy about the dependency inversion here, but this is just a band-aid until we have time to think about a more proper fix.

Basically there's two dynamic "layers" on top of core ostree going on here - rpm-ostree and container images and we need to figure out a general fix.


jmarrero (approved):
lgtm

PR #436, opened 2022-12-11
store: Make failure to query base image non-fatal

cgwalters:
This is another necessary fix for
https://github.com/coreos/rpm-ostree/pull/4204


jmarrero (commented):
lgtm

PR #438, opened 2022-12-11
lib: Switch to olpc-cjson

cgwalters:
This is maintained as part of https://github.com/awslabs/tough and is used by several crates, including the pure Rust OCI crate https://lib.rs/crates/oci-distribution

The immediate motivation is dropping our duplicate `itoa` dep.


jmarrero (approved):
lgtm

PR #439, opened 2022-12-13
container/store: Close image in happy path

cgwalters:
Closes: https://github.com/coreos/rpm-ostree/issues/4213

This fixes leaking the temporary directory but only in the "happy path" i.e. non-error paths for now.

Handling this better would require some API changes to ensure we can `impl Drop` on the image.


jmarrero (approved):
lgtm

PR #440, opened 2022-12-19
container: Fix idempotency of config defaults merging

cgwalters:
In rpm-ostree we provide a custom config to override the isolation user, calling the merge function.  But the logic for idempotency of auth handling was buggy - it needed to consider having `auth_data` already specified.


jmarrero:
Even with the CI failure adding this patch make privileged pulls work again.

cgwalters:
That CI failure was a network flake, restarted it.

cgwalters:
> Even with the CI failure adding this patch make privileged pulls work again.

Can you officially stamp the PR as approved?

jmarrero (approved):
lgtm

PR #442, opened 2023-01-03
isolation: Avoid bug with `--pdeathsig SIGTERM` in Turkish locales

cgwalters:
See https://github.com/coreos/rpm-ostree/issues/4237 for the original report.

Only in locales that use the dotless I case conversion we hit on the fact that util-linux uses `strcasecmp` which is locale-sensitive (It shouldn't be for this, because we're not accepting translated
 signal names.  It should be using an ASCII-only string comparison)

Anyways, we can work around this by omitting the leading `SIG` ourselves.


cgwalters:
Filed https://github.com/util-linux/util-linux/issues/1991 for the underlying bug.

jmarrero (approved):
lgtm

PR #443, opened 2023-01-06
[0.8] Two fixes for new `stable-0.8` branch

cgwalters:
I've created a new `stable-0.8` branch here forked from the last 0.8.9 tag in preparation for backporting fixes to that version which is used by rpm-ostree in rhel8.  These are two minor fixes for that branch to start.

---

Remove test parsing old export

The newer rpm-ostree we ship now no longer supports it.  We
aren't going to change anything in the export path on this
branch, so we don't need it.

---

tests: Update to non-deprecated chrono API

The deprecation was introduced in a new chrono release and
is triggering our clippy lints.

(cherry picked from commit 7875bb33ab37807c1cf3f0ca96297d28c1ea0feb)

---

ci: Retarget for this branch

---


jmarrero (approved):
lgtm

PR #444, opened 2023-01-06
[0.8] backport gc

cgwalters:
Depends https://github.com/ostreedev/ostree-rs-ext/pull/443

This backports the GC work to the (new) 0.8 branch.  Thankfully it was all clean cherry :cherries: picks!

--- 

Remove test parsing old export

The newer rpm-ostree we ship now no longer supports it.  We
aren't going to change anything in the export path on this
branch, so we don't need it.

---

tests: Update to non-deprecated chrono API

The deprecation was introduced in a new chrono release and
is triggering our clippy lints.

(cherry picked from commit 7875bb33ab37807c1cf3f0ca96297d28c1ea0feb)

---

ci: Retarget for this branch

---

container: Prune image layers by default

A while ago we added APIs to do this, but we should really
do it by default when fetching upgraded images.  Otherwise
we effectively leak space.

(cherry picked from commit 97d8d3ded5ff2c02b49d6e3085d9b0fef5feae9d)

---

container: Add an API to query information from a commit object

Prep for better support for pruning.  An ostree deployment
will retain a strong reference solely to a commit object; we can't
rely on anything else.

I plan to use this in rpm-ostree to display metadata information
about the container used for a deployment, even if the
image reference has been pruned.

(cherry picked from commit ee502e54549797161afc57f885c7c4acbc67b07e)

---

container: Add deployed commits into set of GC roots

Prep for handling image pruning better.  The way things
are kind of expected to work today is that for a deployed ostree
commit, we have *two* refs which point to it - one like e.g.
`fedora:fedora/x86_64/coreos/stable`, as well as the "deployment ref"
like "ostree/0/1/1" which is a synthetic ref generated by the
sysroot core.

We want to be able to remove the container image refs - but
doing so today subjects the *layer* branches to garbage collection.

Fix this by looking at the deployment refs as well as the set of
images when computing the set of references for container images.

(cherry picked from commit aec69c8171d797176c795dbd9c2a434b2d5c785e)

---

container: Make single image removal idempotent

Due to bugs or logic errors, it's possible that we might attempt
to prune an image that's already been pruned.  In order to
make things more robust, let's avoid making this a hard error.

Keep the semantics for the existing "prune multiple images" API,
but add a new one that prunes a singular image, and does not
error in the not-found case.  This can be used by higher level
software.

(cherry picked from commit f1af4f34147e43edbcf831198fb09fba0c3c6141)

---

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -7,8 +7,6 @@ set -euo pipefail
 sysroot=/run/host
 # Current stable image fixture
 image=quay.io/fedora/fedora-coreos:testing-devel
-# An unchunked v1 image
-old_image=quay.io/cgwalters/fcos:unchunked
 imgref=ostree-unverified-registry:${image}
 stateroot=testos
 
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

PR #445, opened 2023-01-09
container/store: Write final commit with timestamp

cgwalters:
I was doing some bootc tests and noticed that the deployment commit was different each time I ran an install.  Write the commit with the timestamp of the config (or manifest) to increase reproducibility.


cgwalters:
Draft since I think this works but haven't tested manually, and it could use integration tests.

jmarrero (approved):
/lgtm

PR #447, opened 2023-01-11
two error prefixing patches

cgwalters:
tar: Add error prefixing around commit verification

In an error message we got a bare
`error: No such remote ...`
Let's clarify where this is coming from.

---

store: Add error prefixing on base import

On general principle; we got a "bare errors" in the case of
a missing ostree remote.

---


jmarrero (approved):
lgtm

PR #448, opened 2023-01-11
Add even more error prefixing

cgwalters:
I looked through more of the container and tar paths, and decided to add even more error prefixing to aid future debugging.


jmarrero (approved):
lgtm

Issue #450, opened 2023-01-12
add a version number API


cgwalters:
Today we hardcode looking up and injecting a label key `version`; I did this because it's what's used by e.g. RHEL UBI (and the fedora/centos images).  However...the OCI spec calls instead for `org.opencontainers.image.version` so arguably we should use that.

(That key is used by e.g. `registry.opensuse.org/opensuse/leap:latest`)

But then I was going to look at what `docker.io/library/busybox` does and it's neither?  I guess they just rely on tags-as-versions.

(There is also apparently the [label schema](https://label-schema.org/rc1/) stuff that predated OCI but I am not sure anyone uses that today)

(Except there's a `    "DockerVersion": "20.10.12" ` thing...what the heck is that?  It seems to be on many images on the Docker Hub?)

Edit: `DockerVersion` is something that seems to be injected by `docker build` or so, it's not related to this.

Edit: xref https://github.com/containers/oci-spec-rs/pull/118

PR #451, opened 2023-01-12
container: Add an API to retrieve the version number

cgwalters:
This makes it more convenient for higher level tools to display the associated image version number.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/450


jmarrero (approved):
lgtm

PR #452, opened 2023-01-12
tests: Trim some v0 export format code

cgwalters:
We're not supporting this anymore; continue to trim it from our tests.

(There's even more to do here, but this compiles)


jmarrero (approved):
lgtm

PR #453, opened 2023-01-12
tests: Drop even more export layout bits

cgwalters:
This was all hardcoded to V1, so just drop the conditionals.


jmarrero (approved):
lgtm

PR #454, opened 2023-01-13
container: Add standard OCI version key by default

cgwalters:
And add an API to allow opting-out of the legacy label.


cgwalters:
OK, I reworked this to be less of a flag day.  We'll have both for a while, and allow people to opt-in now to only having the new version label.

jmarrero (approved):
lgtm

PR #456, opened 2023-01-20
Improving the encapsulation (chunking) algorithm 

RishabhSaini:
Followup: [rpm-ostree/issues/4247](https://github.com/coreos/rpm-ostree/issues/4247)
To do:
- [x] Adding package change frequency heuristics
- [x] Adding annotations to oci.layers ot include packing structure
- [x] Adding prior build metadata into basic_packing
- [ ] Modify tests to cover the new algorithm


lib/src/chunking.rs
```diff
@@ -364,78 +380,447 @@ impl Chunking {
...
+                .sum::<f64>()
+                / count as f64;
+            Some(variance.sqrt())
+        }
+        _ => None,
+    }
+}
+
+fn median_absolute_deviation(data: &mut Vec<u64>) -> (f64, f64) {
+    //Sort data
```
cgwalters, line 421: Why the commented out code here?

lib/src/chunking.rs
```diff
@@ -364,78 +380,447 @@ impl Chunking {
...
+    };
+
+    //Absolute deviations
+    let mut absolute_deviations = Vec::new();
+    for size in data {
+        absolute_deviations.push(f64::abs(*size as f64 - median_data))
+    }
+
+    absolute_deviations.sort_by(|a, b| a.partial_cmp(b).unwrap());
+    let mad: f64 = match absolute_deviations.len() % 2 {
```
cgwalters, line 437: Minor, I think this would be a bit more readable to do e.g.
```
let l = absolute_deviations.len();
match l % 2 {
...
```
etc.

lib/src/chunking.rs
```diff
@@ -364,78 +380,447 @@ impl Chunking {
...
+                + absolute_deviations[absolute_deviations.len() / 2])
+        }
+    };
+
+    (median_data, mad)
+}
+
+//Assumes components is sorted by descending size
+//Use MAD as threshold to partition packages [abs(low_limit), high_limit]
+fn get_partitions_with_threshold(
```
cgwalters, line 450: This one could se a doc comment that describes what it does at a high level.
Also in Rust, official "documentation comments" use `///`

cgwalters (commented):
A lot to go over here but not seeing anything major to fix so far; nice work!

lib/src/chunking.rs
RishabhSaini: Because the give data is already sorted in descending order. Sorting a sorted array could be computationally expensive and redundant. 
This leads to the question whether functions should be independent of any assumptions of the given input at the cost of efficiency

lib/src/chunking.rs
```diff
@@ -291,35 +309,33 @@ impl Chunking {
...
+        println!("Time elapsed in packing: {:#?}", duration);
 
         for bin in packing.into_iter() {
-            let first = bin[0];
-            let first_name = &*first.meta.name;
             let name = match bin.len() {
-                0 => unreachable!(),
-                1 => Cow::Borrowed(first_name),
-                2..=5 => {
-                    let r = bin.iter().map(|v| &*v.meta.name).fold(
```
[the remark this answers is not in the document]
RishabhSaini: followup: https://github.com/ostreedev/ostree-rs-ext/pull/472

cgwalters:
This one needs a rebase, right?

cgwalters:
Moving to draft, can you move out of draft when you think it's ready for another review?

cgwalters:
I pushed two commits here, can you take a look to see if they make sense?

lib/src/chunking.rs
```diff
@@ -294,20 +311,27 @@ impl Chunking {
...
 
         // TODO: Compute bin packing in a better way
+        let start = Instant::now();
         let packing = basic_packing(
             sizes,
             NonZeroU32::new(self.max).unwrap(),
             prior_build_metadata,
         );
+        let duration = start.elapsed();
+        println!("Time elapsed in packing: {:#?}", duration);
```
cgwalters, line 321: Let's use something more like `tracing::debug!` for this so it's configurable.  In general, *library* code should not print to stdout unless it's asked I think.

lib/src/chunking.rs
```diff
@@ -372,82 +394,409 @@ impl Chunking {
...
+        else if (freq >= med_freq_high_limit) && (size <= med_size_low_limit) {
+            partitions
+                .entry("hf_ls".to_string())
+                .and_modify(|bin| bin.push(pkg))
+                .or_insert_with(|| vec![pkg]);
+        }
+    }
+
+    for (name, pkgs) in &partitions {
+        println!("{:#?}: {:#?}", name, pkgs.len());
```
cgwalters, line 574: Also `tracing::debug!` here I think.

lib/src/chunking.rs
```diff
@@ -372,82 +394,409 @@ impl Chunking {
...
+        }
+        let mut rem_pkgs_v: Vec<String> = Vec::new();
+        for pkg in removed {
+            rem_pkgs_v.push(pkg.to_string());
+        }
+        let curr_build_len = &curr_build.len();
+        curr_build[curr_build_len - 1].retain(|name| !name.is_empty());
+        curr_build[curr_build_len - 1].extend(add_pkgs_v);
+        for bin in curr_build.iter_mut() {
+            bin.retain(|pkg| !rem_pkgs_v.contains(pkg));
```
cgwalters, line 653: `.contains()` on a `Vec` is O(N)...why are we not just looking up in `removed`?  Oh...are you hitting `&String` versus `&str` issues?

lib/src/chunking.rs
```diff
@@ -372,82 +394,409 @@ impl Chunking {
...
+                prev_pkgs.push(pkg.to_string());
+            }
+        }
+        prev_pkgs.retain(|name| !name.is_empty());
+        let curr_pkgs: Vec<String> = components
+            .iter()
+            .map(|pkg| pkg.meta.name.to_string())
+            .collect();
+        let prev_pkgs_set: HashSet<String> = HashSet::from_iter(prev_pkgs);
+        let curr_pkgs_set: HashSet<String> = HashSet::from_iter(curr_pkgs);
```
cgwalters, line 638: Then we're taking the vector copy and making a `HashSet`?  I think we can just directly do:

```suggestion
        let curr_pkgs_set: HashSet<_> = components.iter().map(|pkg| pkg.meta.name.as_str()).collect();
```

And this will make a set of *borrowed* strings instead of a deep copy.

lib/src/chunking.rs
```diff
@@ -372,82 +394,409 @@ impl Chunking {
...
+            let mut mod_bin = Vec::new();
+            for pkg in bin {
+                mod_bin.push(name_to_component[&pkg]);
+            }
+            modified_build.push(mod_bin);
+        }
+        let mut after_processing_pkgs_len = 0;
+        modified_build.iter().for_each(|bin| {
+            after_processing_pkgs_len += bin.len();
+        });
```
cgwalters, line 672:
```suggestion
        let after_processing_pkgs_len = modified_build.iter().fold(0, |acc, bin| acc + bin.len());
```

or `modified_build.iter().map(|b| b.len()).sum()` which is perhaps clearer?

cgwalters (changes requested):

lib/src/chunking.rs
RishabhSaini: This would lead to the prev_pkgs_set be Vec<String> and the curr_pkgs_set to be Vec<Rc<str>> leading to incosistency

cgwalters: Aren't in both cases we're borrowing a `&str`?

cgwalters: OK this compiles at least, WDYT?

```
diff --git a/lib/src/chunking.rs b/lib/src/chunking.rs
index 18229a51..f8af0241 100644
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -571,21 +571,14 @@ fn basic_packing<'a>(
         let mut curr_build: Vec<Vec<String>> = prior_build.clone();
         //Packing only manaages RPMs not OStree commit
         curr_build.remove(0);
-        let mut prev_pkgs: Vec<String> = Vec::new();
-        for bin in &curr_build {
-            for pkg in bin {
-                prev_pkgs.push(pkg.to_string());
-            }
-        }
-        prev_pkgs.retain(|name| !name.is_empty());
-        let curr_pkgs: Vec<String> = components
+        let prev_pkgs_set: HashSet<_> = curr_build
             .iter()
-            .map(|pkg| pkg.meta.name.to_string())
+            .flat_map(|b| b.iter().map(|p| p.as_str()))
+            .filter(|&name| name != "")
             .collect();
-        let prev_pkgs_set: HashSet<String> = HashSet::from_iter(prev_pkgs);
-        let curr_pkgs_set: HashSet<String> = HashSet::from_iter(curr_pkgs);
-        let added: HashSet<&String> = curr_pkgs_set.difference(&prev_pkgs_set).collect();
-        let removed: HashSet<&String> = prev_pkgs_set.difference(&curr_pkgs_set).collect();
+        let curr_pkgs_set: HashSet<_> = components.iter().map(|pkg| &*pkg.meta.name).collect();
+        let added: HashSet<_> = curr_pkgs_set.difference(&prev_pkgs_set).collect();
+        let removed: HashSet<_> = prev_pkgs_set.difference(&curr_pkgs_set).collect();
         let mut add_pkgs_v: Vec<String> = Vec::new();
         for pkg in added {
             add_pkgs_v.push(pkg.to_string());

```

cgwalters: Note e.g. that `&*pkg.meta.name` lets us go from `Rc<str>` to `&str`.

RishabhSaini: ok makes sense

cgwalters: Did you want to try taking this patch and integrating it?  I think there's some more cleanup along these lines we could do below this too

RishabhSaini: it makes the `removed` type to `HashSet<&&str>` causing problem here:
```
for bin in curr_build.iter_mut() {
            bin.retain(|pkg| !removed.contains(pkg));
        }
```

cgwalters: OK, this is tricky.  I think what we're running up against here is that we're passing around a `Vec<String>` in many places.  A lot of the rest of this code uses `Rc<str>` precisely as a way to share strings in multiple data structures.

I think it'd help to make a type alias.  If we did e.g. 

```
type PackageList = Vec<String>;
```
then it'd be a lot more self describing particularly when used inside a `Vec` itself, i.e. we go from `Vec<Vec<String>>` to `Vec<PackageList>`.

There's another thing here which is that I think in general we don't care about ordering *until* we go to serialize it (in which case it should be sorted).

So we could also make this

```
type PackageSet = HashSet<String>;
```

But, combining all this it'd be

```
type PackageSet = HashSet<Rc<str>>;
```

?

lib/src/container/store.rs
```diff
@@ -204,11 +204,10 @@ pub struct PreparedImport {
 }
 
 impl PreparedImport {
-    /// Iterate over all layers; the ostree split object layers, the commit layer, and any non-ostree layers.
+    /// Iterate over all layers; the commit layer, the ostree split object layers, and any non-ostree layers.
     pub fn all_layers(&self) -> impl Iterator<Item = &ManifestLayerState> {
-        self.ostree_layers
-            .iter()
-            .chain(std::iter::once(&self.ostree_commit_layer))
+        std::iter::once(&self.ostree_commit_layer)
```
cgwalters, line 209: Why this change?

lib/src/chunking.rs
```diff
@@ -469,4 +799,132 @@ mod test {
...
+            .expect("build config descriptor");
+
+        let layers: Vec<oci_spec::image::Descriptor> = metadata_with_ostree_commit
+            .iter()
+            .map(|l| {
+                oci_spec::image::DescriptorBuilder::default()
+                    .media_type(oci_spec::image::MediaType::ImageLayerGzip)
+                    .size(100)
+                    .digest(format!("sha256:{}", l.len()))
+                    .annotations(HashMap::from([("Content".to_string(), l.join(","))]))
```
cgwalters, line 894: Let's make this `"Content"` a `const CONTENT_ANNOTATION: &str = "Content";` above?

lib/src/chunking.rs
```diff
@@ -364,78 +398,374 @@ impl Chunking {
...
+            };
+
+            let pkg_per_bin_ms: usize =
+                (components_len_after_max_freq - limit_hs_bins - limit_ls_pkgs)
+                    .checked_div(limit_ms_bins)
+                    .expect("number of bins should be >= 4");
+
+            //Bins assignment
+            for partition in partitions.keys() {
+                let pkgs = partitions.get(partition).expect("hashset");
```
cgwalters, line 693: I think this is just
```suggestion
            for (partition, pkgs) in partitions.iter() {
```
right?

lib/src/chunking.rs
```diff
@@ -364,78 +398,374 @@ impl Chunking {
...
+    //  iterate through prior[i] and make bins according to the name in nevra of pkgs and return
+    //  (no need of recomputing packaging structure)
+    //else if pkg structure to be changed || prior build not specified
+    //  Recompute optimal packaging strcuture (Compute partitions, place packages and optimize build)
+
+    if let Some(prior_build) = prior_build_packages
+    /* && structure not be changed*/
+    {
+        tracing::debug!("Keeping old package structure");
+        let mut curr_build: Vec<Vec<String>> = prior_build.clone();
```
cgwalters, line 588: As the linter is saying, we can drop this clone now

cgwalters (commented):
A few quick comments.  Thanks for adding the test case, that should really help with future changes!

cgwalters:
WDYT about this change?

```
diff --git a/lib/src/chunking.rs b/lib/src/chunking.rs
index 6f69d709..c4efb93c 100644
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -599,17 +599,18 @@ fn basic_packing<'a>(
             .iter()
             .map(|pkg| pkg.meta.name.to_string())
             .collect();
+        // View the packages as unordered sets for lookups and differencing.
         let prev_pkgs_set: HashSet<String> = HashSet::from_iter(prev_pkgs);
         let curr_pkgs_set: HashSet<String> = HashSet::from_iter(curr_pkgs);
-        let added: HashSet<&String> = curr_pkgs_set.difference(&prev_pkgs_set).collect();
-        let removed: HashSet<&String> = prev_pkgs_set.difference(&curr_pkgs_set).collect();
-        let mut add_pkgs_v: Vec<String> = Vec::new();
-        for pkg in added {
-            add_pkgs_v.push(pkg.to_string());
+
+        // Add all new packages into the last bin, which was reserved as empty space.
+        if let Some(last_bin) = curr_build.last_mut() {
+            let added = curr_pkgs_set.difference(&prev_pkgs_set);
+            last_bin.retain(|name| !name.is_empty());
+            last_bin.extend(added.into_iter().cloned());
         }
-        let curr_build_len = &curr_build.len();
-        curr_build[curr_build_len - 1].retain(|name| !name.is_empty());
-        curr_build[curr_build_len - 1].extend(add_pkgs_v);
+
+        let removed: HashSet<&String> = prev_pkgs_set.difference(&curr_pkgs_set).collect();
         for bin in curr_build.iter_mut() {
             bin.retain(|pkg| !removed.contains(pkg));
         }
```

- Use `last_mut()` to get access to the last element of the vector in a safer way (we avoid panicking if the vec is somehow zero sized)
- Notice we don't need to directly call `.collect()` to create an intermediate `HashMap` for the set of added packages; the `.difference()` API returns an iterator which we can pass straight to `.extend()`!
- Relatedly, because `added` is only necessary in that section, move it into the conditional block.  There's a *lot* of variables here, and it helps understanding if we reduce the number that are "in scope" at the same time
- Add a few more comments

lib/src/container/store.rs
RishabhSaini: This is fixing a bug that existed before. Essentially  the way we are storing the oci images is "first is ostree_commit_layer, then layers with packages", but when we were doing `layers_to_fetch` the order in `all_layers` was reversed causing wrong asserts to be tested here https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/tests/it/main.rs#L728

cgwalters: Hmmm.  Can you break that out as a separate change in a new PR?  The tests are passing with the current code, right?  So I assume we need a change to the tests to show the failure or so?

RishabhSaini: Followup here:
https://github.com/ostreedev/ostree-rs-ext/issues/480
https://github.com/ostreedev/ostree-rs-ext/pull/481

lib/src/chunking.rs
RishabhSaini: This leads to a borrowed reference when defining sets to compute set differences on and causes errors

lib/src/chunking.rs
```diff
@@ -364,78 +400,369 @@ impl Chunking {
...
+        //1st layer is skipped as packing doesn't manage ostree_commit layer
+        let mut curr_build: Vec<Vec<String>> = prior_build
+            .layers()
+            .iter()
+            .skip(1)
+            .map(|layer| {
+                let annotation_layer = layer
+                    .annotations()
+                    .as_ref()
+                    .expect("Layer does not consist annotation");
```
cgwalters, line 579: I think instead of panic'ing an explicit error would probably be better here.  This would come up in a real world scenario I think where a build system passes a prior build that was generated *before* this code landed.

So something more like:

```suggestion
                let annotation_pkgs = layer
                    .annotations()
                    .as_ref()
                    .and_then(|annos|  annos.get(CONTENT_ANNOTATION))
                    .ok_or_else(|| anyhow!("Prior build is missing annotation {CONTENT_ANNOTATION}");
```

cgwalters:
Does this make sense to you?

```
From 1e3e5b4202d4256e89d1b7381469f953792dcef0 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Mon, 15 May 2023 11:18:37 -0400
Subject: [PATCH] Take prior manifest as reference

Having this as a separate argument instead of part of
ExportOpts allows us to do this.
---
 lib/src/chunking.rs              | 10 +++++-----
 lib/src/container/encapsulate.rs |  6 +++---
 2 files changed, 8 insertions(+), 8 deletions(-)

diff --git a/lib/src/chunking.rs b/lib/src/chunking.rs
index 741cb5e2..43cd758d 100644
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -268,7 +268,7 @@ impl Chunking {
         rev: &str,
         meta: ObjectMetaSized,
         max_layers: &Option<NonZeroU32>,
-        prior_build_metadata: Option<oci_spec::image::ImageManifest>,
+        prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
     ) -> Result<Self> {
         let mut r = Self::new(repo, rev)?;
         r.process_mapping(meta, max_layers, prior_build_metadata)?;
@@ -286,7 +286,7 @@ impl Chunking {
         &mut self,
         meta: ObjectMetaSized,
         max_layers: &Option<NonZeroU32>,
-        prior_build_metadata: Option<oci_spec::image::ImageManifest>,
+        prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
     ) -> Result<()> {
         self.max = max_layers
             .unwrap_or(NonZeroU32::new(MAX_CHUNKS).unwrap())
@@ -547,7 +547,7 @@ fn get_partitions_with_threshold(
 fn basic_packing<'a>(
     components: &'a [ObjectSourceMetaSized],
     bin_size: NonZeroU32,
-    prior_build_metadata: Option<oci_spec::image::ImageManifest>,
+    prior_build_metadata: Option<&oci_spec::image::ImageManifest>,
 ) -> Result<Vec<Vec<&'a ObjectSourceMetaSized>>> {
     let mut r = Vec::new();
     let mut components: Vec<_> = components.iter().collect();
@@ -914,7 +914,7 @@ mod test {
         let packing_derived = basic_packing(
             &contentmeta_v1.as_slice(),
             NonZeroU32::new(6).unwrap(),
-            Some(image_manifest_v0),
+            Some(&image_manifest_v0),
         )
         .unwrap();
         let structure_derived: Vec<Vec<&str>> = packing_derived
@@ -953,7 +953,7 @@ mod test {
         let packing_derived = basic_packing(
             &contentmeta_v2.as_slice(),
             NonZeroU32::new(6).unwrap(),
-            Some(image_manifest_v1),
+            Some(&image_manifest_v1),
         )
         .unwrap();
         let structure_derived: Vec<Vec<&str>> = packing_derived
diff --git a/lib/src/container/encapsulate.rs b/lib/src/container/encapsulate.rs
index 66918656..742c68c6 100644
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -191,7 +191,7 @@ fn build_oci(
     tag: Option<&str>,
     config: &Config,
     opts: ExportOpts,
-    prior_build: Option<oci_image::ImageManifest>,
+    prior_build: Option<&oci_image::ImageManifest>,
     contentmeta: Option<crate::chunking::ObjectMetaSized>,
 ) -> Result<ImageReference> {
     if !ocidir_path.exists() {
@@ -317,7 +317,7 @@ async fn build_impl(
     repo: &ostree::Repo,
     ostree_ref: &str,
     config: &Config,
-    prior_build: Option<oci_image::ImageManifest>,
+    prior_build: Option<&oci_image::ImageManifest>,
     opts: Option<ExportOpts>,
     contentmeta: Option<ObjectMetaSized>,
     dest: &ImageReference,
@@ -406,7 +406,7 @@ pub async fn encapsulate<S: AsRef<str>>(
     repo: &ostree::Repo,
     ostree_ref: S,
     config: &Config,
-    prior_build: Option<oci_image::ImageManifest>,
+    prior_build: Option<&oci_image::ImageManifest>,
     opts: Option<ExportOpts>,
     contentmeta: Option<ObjectMetaSized>,
     dest: &ImageReference,
-- 
2.39.0

```
?

cgwalters:
Also

```
From 1c847e578c3be47e4de4fe19fde66149b2cb8c52 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Mon, 15 May 2023 12:44:26 -0400
Subject: [PATCH] chunking: Avoid intermediate Vec allocation

We can directly create the set from an iterator, same as
`curr_pkgs_set`.
---
 lib/src/chunking.rs | 12 ++++++------
 1 file changed, 6 insertions(+), 6 deletions(-)

diff --git a/lib/src/chunking.rs b/lib/src/chunking.rs
index 43cd758d..8a6fe9f8 100644
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -583,12 +583,12 @@ fn basic_packing<'a>(
             .collect();
         let mut curr_build = curr_build?;
 
-        //Flatten and filter
-        let mut prev_pkgs: Vec<String> = curr_build.concat();
-        prev_pkgs.retain(|name| !name.is_empty());
-
-        //View the packages as unordered sets for lookups and differencing
-        let prev_pkgs_set: HashSet<String> = HashSet::from_iter(prev_pkgs);
+        // View the packages as unordered sets for lookups and differencing
+        let prev_pkgs_set: HashSet<String> = curr_build
+            .iter()
+            .flat_map(|v| v.iter().cloned())
+            .filter(|name| !name.is_empty())
+            .collect();
         let curr_pkgs_set: HashSet<String> = components
             .iter()
             .map(|pkg| pkg.meta.name.to_string())
-- 
2.39.0

```
?

cgwalters:
Pushed a few clippy lint fixes

RishabhSaini:
The changes look good to me. Thanks for the help!

RishabhSaini:
There is an linting error: 
```
error: unknown lint: `clippy::bool_to_int_with_if`
```

cgwalters:
Ah fun, it's a new lint.  OK, I tried another tack for that one.

cgwalters:
How about this?

```
From 8e9abe479775e81cdf92e218cd28a1cf16f52e99 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Mon, 15 May 2023 14:46:17 -0400
Subject: [PATCH] Use CONTENT_ANNOTATION in more places, change name

Use the `const` we defined so it's easy to find the uses of it.
Also, change the name to be "namespaced" more clearly;
see https://github.com/opencontainers/image-spec/blob/main/annotations.md
for more.
---
 lib/src/chunking.rs              | 2 +-
 lib/src/container/encapsulate.rs | 9 ++++++---
 lib/src/container/mod.rs         | 4 ++++
 3 files changed, 11 insertions(+), 4 deletions(-)

diff --git a/lib/src/chunking.rs b/lib/src/chunking.rs
index 6d77e23a..c8fdc333 100644
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -10,6 +10,7 @@ use std::num::NonZeroU32;
 use std::rc::Rc;
 use std::time::Instant;
 
+use crate::container::CONTENT_ANNOTATION;
 use crate::objectsource::{ContentID, ObjectMeta, ObjectMetaMap, ObjectSourceMeta};
 use crate::objgv::*;
 use crate::statistics;
@@ -32,7 +33,6 @@ pub(crate) type ChunkMapping = BTreeMap<RcStr, (u64, Vec<Utf8PathBuf>)>;
 
 const LOW_PARTITION: &str = "2ls";
 const HIGH_PARTITION: &str = "1hs";
-const CONTENT_ANNOTATION: &str = "Content";
 
 #[derive(Debug, Default)]
 pub(crate) struct Chunk {
diff --git a/lib/src/container/encapsulate.rs b/lib/src/container/encapsulate.rs
index 4a9b7cd7..ca09c61a 100644
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -1,7 +1,7 @@
 //! APIs for creating container images from OSTree commits
 
 use super::ocidir::{Layer, OciDir};
-use super::{ocidir, OstreeImageReference, Transport};
+use super::{ocidir, OstreeImageReference, Transport, CONTENT_ANNOTATION};
 use super::{ImageReference, SignatureSource, OSTREE_COMMIT_LABEL};
 use crate::chunking::{Chunk, Chunking, ObjectMetaSized};
 use crate::container::skopeo;
@@ -152,7 +152,10 @@ fn export_chunked(
 
     // Add the ostree layer
     let mut annotation_ostree_layer = HashMap::new();
-    annotation_ostree_layer.insert("Content".to_string(), "ostree_commit".to_string());
+    annotation_ostree_layer.insert(
+        super::CONTENT_ANNOTATION.to_string(),
+        "ostree_commit".to_string(),
+    );
     ociw.push_layer(
         manifest,
         imgcfg,
@@ -163,7 +166,7 @@ fn export_chunked(
     // Add the component/content layers
     for (layer, name, packages) in layers {
         let mut annotation_component_layer = HashMap::new();
-        annotation_component_layer.insert("Content".to_string(), packages.join(","));
+        annotation_component_layer.insert(CONTENT_ANNOTATION.to_string(), packages.join(","));
         ociw.push_layer(
             manifest,
             imgcfg,
diff --git a/lib/src/container/mod.rs b/lib/src/container/mod.rs
index e2bb7970..115912ca 100644
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -37,6 +37,10 @@ use std::str::FromStr;
 /// The label injected into a container image that contains the ostree commit SHA-256.
 pub const OSTREE_COMMIT_LABEL: &str = "ostree.commit";
 
+/// The name of an annotation attached to a layer which names the packages/components
+/// which are part of it.
+pub(crate) const CONTENT_ANNOTATION: &str = "ostree.components";
+
 /// Our generic catchall fatal error, expected to be converted
 /// to a string to output to a terminal or logs.
 type Result<T> = anyhow::Result<T>;
-- 
2.39.0

```
?

cgwalters:
Do we actually need the content annotation `ostree_commit` at all though?

RishabhSaini:
No, it was inserted just for consistency and readability

cgwalters:
OK hmm.  I think I'd say let's drop it because with it present, we're "mixing namespaces" - the things we put in here are otherwise strings provided externally.  So if one was writing a tool to process this data and map it back to package names, they'd need to skip the layer.  Which is what we do anyways too!  So I don't see an advantage to adding it.

cgwalters (approved):
OK let's land this and try it out and make further improvements after merging!

cgwalters:
Although, I think it's probably cleaner to squash everything down to one commit; WDYT?

RishabhSaini:
yup squashing the commits

cgwalters:
OK following up here related to [this comment](https://github.com/coreos/rpm-ostree/pull/4271#issuecomment-1550146161).  Basically a core tradeoff here is that in the case where a prior build is *not* used as a base, this causes the packing to be *much* more sensitive to the input.  And the consequence then is we'll actually cause *more* data to be downloaded in many cases.

For the purposes of gathering some real-world data, I used the scripts linked above to construct two chains of images at `quay.io/cgwalters/fcos:b{1..15}` `quay.io/cgwalters/fcos:br{1..15}` (`b` here meaning "build", and `r` expanding to "repacked").

# Before this PR

```
$ for x in $(seq 14); do echo $x '->' $(($x + 1)); ostree container compare ostree-unverified-registry:quay.io/cgwalters/fcos:b$x ostree-unverified-registry:quay.io/cgwalters/fcos:b$(($x + 1)); echo ;done
1 -> 2
Total new layers: 51    Size: 730.1 MB
Removed layers:   28    Size: 525.0 MB
Added layers:     28    Size: 540.4 MB

2 -> 3
Total new layers: 51    Size: 730.8 MB
Removed layers:   46    Size: 567.6 MB
Added layers:     46    Size: 568.3 MB

3 -> 4
Total new layers: 51    Size: 727.4 MB
Removed layers:   13    Size: 316.9 MB
Added layers:     13    Size: 313.6 MB

4 -> 5
Total new layers: 51    Size: 711.7 MB
Removed layers:   24    Size: 578.8 MB
Added layers:     24    Size: 563.1 MB

5 -> 6
Total new layers: 51    Size: 708.1 MB
Removed layers:   17    Size: 345.9 MB
Added layers:     17    Size: 342.4 MB

6 -> 7
Total new layers: 51    Size: 701.9 MB
Removed layers:   18    Size: 493.4 MB
Added layers:     18    Size: 487.1 MB

7 -> 8
Total new layers: 51    Size: 698.2 MB
Removed layers:   16    Size: 535.2 MB
Added layers:     16    Size: 531.5 MB

8 -> 9
Total new layers: 51    Size: 697.0 MB
Removed layers:   14    Size: 347.5 MB
Added layers:     14    Size: 346.4 MB

9 -> 10
Total new layers: 51    Size: 697.0 MB
Removed layers:   4     Size: 170.5 MB
Added layers:     4     Size: 170.5 MB

10 -> 11
Total new layers: 51    Size: 695.6 MB
Removed layers:   18    Size: 404.9 MB
Added layers:     18    Size: 403.4 MB

11 -> 12
Total new layers: 51    Size: 693.9 MB
Removed layers:   18    Size: 505.2 MB
Added layers:     18    Size: 503.4 MB

12 -> 13
Total new layers: 51    Size: 693.9 MB
Removed layers:   7     Size: 283.9 MB
Added layers:     7     Size: 283.9 MB

13 -> 14
Total new layers: 51    Size: 755.0 MB
Removed layers:   51    Size: 693.9 MB
Added layers:     51    Size: 755.0 MB

14 -> 15
Total new layers: 51    Size: 755.0 MB
Removed layers:   4     Size: 169.3 MB
Added layers:     4     Size: 169.3 MB
$
```

So a user following all those updates would download ~5978 MB.

# After this PR

```
$ for x in $(seq 14); do echo $x '->' $(($x + 1)); ostree container compare ostree-unverified-registry:quay.io/cgwalters/fcos:br$x ostree-unverified-registry:quay.io/cgwalters/fcos:br$(($x + 1)); echo ;done
1 -> 2
Total new layers: 65    Size: 729.4 MB
Removed layers:   48    Size: 544.5 MB
Added layers:     48    Size: 560.2 MB

2 -> 3
Total new layers: 65    Size: 730.0 MB
Removed layers:   60    Size: 572.1 MB
Added layers:     60    Size: 572.6 MB

3 -> 4
Total new layers: 65    Size: 721.4 MB
Removed layers:   35    Size: 340.5 MB
Added layers:     35    Size: 331.9 MB

4 -> 5
Total new layers: 65    Size: 711.0 MB
Removed layers:   44    Size: 586.2 MB
Added layers:     44    Size: 575.7 MB

5 -> 6
Total new layers: 65    Size: 707.4 MB
Removed layers:   32    Size: 342.5 MB
Added layers:     32    Size: 339.0 MB

6 -> 7
Total new layers: 65    Size: 701.3 MB
Removed layers:   37    Size: 505.2 MB
Added layers:     37    Size: 499.1 MB

7 -> 8
Total new layers: 65    Size: 697.6 MB
Removed layers:   33    Size: 530.3 MB
Added layers:     33    Size: 526.6 MB

8 -> 9
Total new layers: 65    Size: 713.8 MB
Removed layers:   63    Size: 694.1 MB
Added layers:     63    Size: 710.3 MB

9 -> 10
Total new layers: 65    Size: 696.5 MB
Removed layers:   63    Size: 710.3 MB
Added layers:     63    Size: 693.1 MB

10 -> 11
Total new layers: 65    Size: 695.1 MB
Removed layers:   34    Size: 410.1 MB
Added layers:     34    Size: 408.6 MB

11 -> 12
Total new layers: 65    Size: 693.5 MB
Removed layers:   37    Size: 522.7 MB
Added layers:     37    Size: 521.1 MB

12 -> 13
Total new layers: 65    Size: 693.5 MB
Removed layers:   13    Size: 228.6 MB
Added layers:     13    Size: 228.6 MB

13 -> 14
Total new layers: 65    Size: 754.6 MB
Removed layers:   43    Size: 558.2 MB
Added layers:     43    Size: 619.3 MB

14 -> 15
Total new layers: 65    Size: 754.6 MB
Removed layers:   3     Size: 101.8 MB
Added layers:     3     Size: 101.8 MB
$
```

Whereas after, they'd download ~6687MB, a roughly 11% increase.  

However it's not quite the average that concerns me so much as the outliers here; for example, update 9 :arrow_right: 10 goes from 170.5 to 693.1.

But...maybe we just live with this and aim to push users to generally pin to prior builds.

cgwalters:
Oh.  I think something went wrong here...b9 and b10 are different things when repacked.

RishabhSaini:
What are the fcos version numbers for b9 and b10?

cgwalters:
Yeah, the versions are different:

```
$ skopeo inspect -n docker://quay.io/cgwalters/fcos:br9 | grep -i version
    "DockerVersion": "",
        "org.opencontainers.image.version": "38.20230430.3.0",
        "version": "38.20230430.3.0"
$ skopeo inspect -n docker://quay.io/cgwalters/fcos:b9 | grep -i version
    "DockerVersion": "",
        "version": "37.20230110.3.1"
```

So something went wrong with my scripting...looking

cgwalters:
OK I think a core bug here may relate to `skopeo copy ... oci:` *not removing previous tags*.  I will drill a bit more into that to be sure but I've now cleaned out things and rerun my scripts again and posted [updated numbers](https://github.com/ostreedev/ostree-rs-ext/pull/456#issuecomment-1551463117).  I think we're good now!

RishabhSaini:
Defaulting to using the `--previous-build-manifest` should give better numbers. Computing the packing structure every single time is not what the algorithm intended. It should only be done once every major release of the OS

cgwalters:
Right.  I'll look at doing more numbers with that enabled.  But I do think we still need to care about the no-previous-state case; if nothing else, it's what will happen if we ship this code today *without* updating coreos-assembler e.g. too.

Until now users have been able to rely on the "default" rpm-ostree chunking.

PR #457, opened 2023-02-08
container: Add a `copy_as` function, deprecate `copy`

cgwalters:
For use cases like installers where we want to pull from a source like `containers-storage:` or `oci:`, but have the destination ostree repo be configured to fetch from `docker://` AKA `registry:`.

The deploy/import code already supports this via
the `target_imgref` flow.  But I have a new code path in bootc where I want to copy from one ostree repo to another, while also performing this conversion.

While we're here, also ensure this API operates in terms of `ImageReference`, because that's all
the ostree storage uses.

And finally, deprecate `copy` because of the API
wart above; `copy_as` is more general and not
much more typing in the simple case of wanting
to retain the same image reference.


jmarrero (approved):
lgtm

PR #458, opened 2023-02-09
Add ostree-ext-cli container compare image1 image2

RishabhSaini:
Allows to compare the content between two OCI compliant images

### DIFF
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -27,6 +27,7 @@
 
 use anyhow::anyhow;
 
+use ostree::glib;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ops::Deref;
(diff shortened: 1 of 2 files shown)


lib/src/container/mod.rs
```diff
@@ -246,6 +247,64 @@ impl std::fmt::Display for OstreeImageReference {
...
+        if !src_layers.contains_key(blobid) {
+            diff.added.push(descriptor.clone());
+        }
+    }
+    diff
+}
+
+impl ManifestDiff {
+    /// Prints the total, removed and added content between two OCI images
+    pub fn print(&self) {
```
cgwalters, line 290: I go back and forth on whether we should have code that calls `println!` in the library.  In some ways it feels architecturally unclean, but in practice it's...practical :smile:

cgwalters (approved):
Nice!

cgwalters:
BTW though it would be nice to have some unit/integration tests for this.  Maybe in `priv-integration.sh` to say diff an image with itself and verify it shows zero modified layers to start?

lib/src/container/mod.rs
RishabhSaini: agreed

RishabhSaini:
> BTW though it would be nice to have some unit/integration tests for this. Maybe in `priv-integration.sh` to say diff an image with itself and verify it shows zero modified layers to start?

https://github.com/ostreedev/ostree-rs-ext/pull/459

PR #462, opened 2023-03-01
deploy: Make `--stateroot` default to `default`

cgwalters:
See https://github.com/ostreedev/ostree/issues/2794

Having this e.g. be `rhcos` or `fcos` etc. doesn't add any value.


jmarrero (approved):
lgtm

PR #464, opened 2023-03-03
deploy: Add optional `--image` syntax

cgwalters:
This clones the nicer bootc syntax which does away with forcing users to understand the image reference strings.

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -23,11 +23,19 @@ fi
 if test '!' -d "${sysroot}/ostree/deploy/${stateroot}"; then
     ostree admin os-init "${stateroot}" --sysroot "${sysroot}"
 fi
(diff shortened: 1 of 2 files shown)


jkonecny12:
Thanks a lot for this simplification!

cgwalters:
To xref this is to aid https://github.com/rhinstaller/anaconda/pull/4561 - specifically anaconda's kickstart syntax can just be a literal passthrough for these options, i.e. we'd support
`ostreecontainer --image quay.io/exampleuser/someimage:latest`
and `ostreecontainer --image quay.io/exampleuser/someimage:latest --no-signature-verification`
etc. without forcing users to understand the arcane syntax of ostree-imgrefs.

jmarrero (approved):
lgtm

PR #465, opened 2023-03-05
container/store: Add a well-known ref for holding base images

cgwalters:
This will allow us to not hardcode rpm-ostree in the future, and will be more extensible in general.  This is targeted for use in bootc.


jmarrero (approved):
lgtm

Issue #468, opened 2023-03-23
commit: Don't remove `/var/tmp`


cgwalters:
OK we've designed a conflicting system; the `ostree container commit` command today will nuke `/var`, but in most implementations this will create a whiteout, which the deploy command warns about.

Also, too many commands expect `/var/tmp` to exist; so creating *further* derived images after `ostree container commit` is problematic.

Instead, let's cave here and prune everything except `/var/tmp`, and also change the deploy code to not warn about empty directories there.

PR #469, opened 2023-03-25
commit: Don't prune `/var`

cgwalters:
We've designed a conflicting system; the `ostree container commit` command today will nuke `/var`, but in most implementations this will create a whiteout, which the deploy command warns about.

Also, too many commands expect `/var/tmp` to exist; so creating *further* derived images after `ostree container commit` is problematic.

Instead, let's cave here and prune everything except `/var/tmp`, and also change the deploy code to not warn about empty directories there.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/468


jmarrero (approved):
/lgtm

PR #471, opened 2023-05-01
Use `oci_spec` re-exported from `containers_image_proxy`

cgwalters:
This needs to stay in sync, because it's a public API dependency of that project.  Motivated by a recent semver bump there.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -11,7 +11,7 @@ rust-version = "1.64.0"
 
 [dependencies]
 anyhow = "1.0"
-containers-image-proxy = "0.5.2"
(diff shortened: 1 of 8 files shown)


cgwalters:
This fixes bugs like https://github.com/coreos/rpm-ostree/pull/4313

cgwalters:
In other words in the end for us, the one thing referencing the specific oci_spec version will be https://github.com/containers/containers-image-proxy-rs/blob/217a614523e5fe442dac37ec0301a04b83a04de8/Cargo.toml#L17

jmarrero (approved):

PR #472, opened 2023-05-02
chunking: Deduplicate the config history for each layer

RishabhSaini:
Followup: https://github.com/ostreedev/ostree-rs-ext/pull/456#discussion_r1180747811
Fix the the repeated name of the first package in the history of oci config for each oci layer


cgwalters (approved):
Thanks!

PR #473, opened 2023-05-08
tests: Port to xshell

cgwalters:
This is better than my own `sh-inline` crate:

- It properly supports variable capture from the Rust context which is way more ergonomic
- There's higher level helper functions too that allow reading/writing files relative to the context directory
- It doesn't depend on cap-std-ext, and we need to bump the semver there
- It's maintained by someone else
- It has more users


jmarrero (approved):
lgtm

PR #474, opened 2023-05-09
Bump to cap-std-ext 2.0, use rustix directly

cgwalters:
In cap-std-ext we made the mistake of making rustix a public API; the 2.0 version fixes that.

Add rustix directly here and use it.


jmarrero (approved):
lgtm

PR #475, opened 2023-05-10
Drop unused format-version API

cgwalters:
Since we're breaking semver we can just drop this.


jmarrero (approved):
lgtm

jmarrero:
@cgwalters and I discussed the trade offs between merging as is as or completely removing the `ExportOptions` code. I was wondering why we need to add the `#[allow(dead_code)]` annotation. Colin explained that now that we are just using format `1`, nothing is using `self.options.format_version`. This raises the idea of completely removing the code instead of annotating it. While completely removing it would be cleaner, we decided to keep it as is for now and will reconsider further down the line if there is no near future need to extend this code.

PR #477, opened 2023-05-10
Drop container ExportLayout from public interface

cgwalters:
Since we're making a semver-incompatible API change, and we don't support V0 anymore, drop the public API touchpoints here.


jmarrero (approved):
lgtm

PR #478, opened 2023-05-10
container: Move ManifestDiff to use a constructor

cgwalters:
In Rust the norm is to use a `fn new` for these types of things.

Just going over our public API and looking for improvements.


jmarrero (approved):
lgtm

PR #479, opened 2023-05-10
tree-wide: Run cargo clippy --fix

cgwalters:
tests: One run of `cargo clippy --fix`

Nothing important, just some unnecessary `&` churn.

---

lib: Ignore incorrect clippy lint about deref+`Lazy`

`cargo clippy --fix` fails to compile if we remove the `*` here.

---

tree-wide: Run `cargo clippy --fix`

Nothing important here, just the usual spam of unnecessary `&`
basically.

---


jmarrero (approved):
lgtm

Issue #480, opened 2023-05-13
test_container_chunked is wrong 


RishabhSaini:
In short: 

According to the chunked container image generated from the Package Database defined [here](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/tests/it/main.rs#L662)

| `ImageConfiguration` lists the layers as | `ImageManifest` has the following SHA256 |
| ----------- | ----------- |
|ostree export of commit 38ab1f9... |5b249...|
|pkgdb | 5a35c...|
|initramfs | d4bc...|
|kernel | ed7d...|
|a-polkit-config | ebbd...|
|someconfig | 970e...|
|testlink | bd089...|
|bash|9bb66...|

While the [`layers_with_history()`](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/src/container/store.rs#L227) combines them erroneously as:

| Pkg | SHA256 of the layer as in `ImageManifest` |
| ----------- | ----------- |
|ostree export of commit 38ab1f9... | 5a35c...|
|pkgdb|d4bc...|
|initramfs | ed7d...|
|kernel | ebbd...|
|a-polkit-config | 970e...|
|someconfig | bd0899...|
|testlink | 9bb66...|
|bash | 5b249...|

This is because [`all_layers()`](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/src/container/store.rs#L208) has `ostree_layers` before `commit_layer` which is not the order in `ImageManifest` or `ImageConfig` history.

The testing is wrong as well [here](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/tests/it/main.rs#LL728C5-L728C5). The layer with `testlink` is never modified and hence should not be updated

In long: 

This is the config history:
```
config: ImageConfiguration {
        created: None,
        author: None,
        architecture: Amd64,
        os: Linux,
        os_version: None,
        os_features: None,
        variant: None,
        config: Some(
            Config {
                user: None,
                exposed_ports: None,
                env: None,
                entrypoint: None,
                cmd: Some(
                    [
                        "/usr/bin/bash",
                    ],
                ),
                volumes: None,
                working_dir: None,
                labels: Some(
                    {
                        "ostree.final-diffid": "sha256:e4df77a4523bb1d92d43ab0ab06e2ed6c43b7192132dff7f0a77fa9350d5ec14",
                        "test": "value",
                        "version": "42.0",
                        "org.opencontainers.image.version": "42.0",
[...]

PR #482, opened 2023-05-15
Two minor test additions

cgwalters:
tests: Also verify manifest digest

Just a drive by cleanup; unused variables are usually a red flag.
Let's just add an assertion here.

---

tests: Add a test for layers_with_history()

Motivated by https://github.com/ostreedev/ostree-rs-ext/issues/480

---


jmarrero (approved):
lgtm

PR #484, opened 2023-05-15
chunking: Minor comment cleanup/clarifications

cgwalters:
The default Rust style includes a space between the comment `//` and the text, so do that consistently.

Ensure that "documentation comments" are consistently associated with their function.

Expand acronyms like "hs" and "ms" etc. in many places because I find this easier to read.  (Particularly I have a strong association with "ms" as "millisecond")

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -411,18 +411,17 @@ fn packing_size(packing: &[Vec<&ObjectSourceMetaSized>]) -> u64 {
     packing.iter().map(|v| components_size(v)).sum()
 }
 
-///Given a certain threshold, divide a list of packages into all combinations
-///of (high, medium, low) size and (high,medium,low) using the following
-///outlier detection methods:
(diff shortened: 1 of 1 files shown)


cgwalters:
@RishabhSaini can you review?

RishabhSaini (approved):
LGTM

cgwalters:
Ahh wait, because https://github.com/ostreedev/ostree-rs-ext/pull/456 wasn't rebased on main, we have a semantic conflict with the new test in https://github.com/ostreedev/ostree-rs-ext/pull/482

```
---- test_container_chunked stdout ----
thread 'test_container_chunked' panicked at 'called `Option::unwrap()` on a `None` value', lib/tests/it/main.rs:709:14
```

It looks like this is because the new packing here notices that everything is tiny?  Looking...

cgwalters:
OK, for now pushed a change to adjust the test.

PR #485, opened 2023-05-17
container: Correctly print size of added layers

cgwalters:
I noticed the sizes were always exactly the same...


cgwalters:
Based this on https://github.com/ostreedev/ostree-rs-ext/pull/486

RishabhSaini:
In priv-integration.sh we need the extra padding adding in the grep statements
```
ostree-ext-cli container compare ${imgref} ${imgref} > compare.txt
grep "Removed layers: 0  Size: 0 bytes" compare.txt
grep "Added layers: 0  Size: 0 bytes" compare.txt
```

cgwalters:
OK, fixed the integration test.

PR #486, opened 2023-05-17
tests: Drop legacy image test

cgwalters:
I cleaned up my quay.io namespace and obviously didn't realize/remember that we were using an image from my namespace in our tests.

While we could recreate it, it doesn't seem worth it.  Just drop the test.


RishabhSaini (approved):
LGTM!

Issue #420, opened 2022-11-23
Encapsulated OSTree commits should retain timestamp


jlebon:
Right now, OSTree containers have:

```
"Created": "0001-01-01T00:00:00Z"
```

IIUC, I think this is because we're not actually setting the `Created` field on [the image configuration](https://github.com/ostreedev/ostree-rs-ext/blob/fa59f819ae1ce5837ada346338a73d56297d757e/lib/src/container/encapsulate.rs#L223).

Should we be setting that to the timestamp of the OSTree commit?

cgwalters:
Hmm, yes, arguably.  That said...I think we should also teach rpm-ostree to probably default to setting a commit timestamp to be the latest timestamp of all input RPMs.  (And for ostree layers, all input layers).

That'd help ensure that generated container images are more reproducible too after we do this.

cgwalters:
https://github.com/ostreedev/ostree-rs-ext/pull/445 is related to this

PR #487, opened 2023-05-17
container: Set creation date to match ostree commit timestamp

cgwalters:
This is just friendlier because we want the value to be potentially useful.  It doesn't harm reproducibility because if the embedded commit timestamp is not reproducible, then the image will already not be reproducible.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/420


jmarrero (approved):
lgtm

jmarrero (approved):
lgtm

PR #489, opened 2023-05-17
A few more minor chunking patches

cgwalters:
chunking: Use `.partition()`

It's made for this situation.

---

chunking: Have partitioning borrow component vec

Prep for further changes.

---

chunking: use `is_empty()` instead of matching on Lines

Main motivation is that this allows dropping a level of indentation,
but it's also more idiomatic I think.

---

chunking: Extract a `const` for high size

This is a bit more self-documenting.

---

chunking: Factor out a common variable for bin computation

To reduce duplication.

---

chunking: Use combinators to compute low size

In this case, the combinator is perhaps mildly easier to read.
Also add comments.

---

chunking: Inline empty bin allocation

And add a comment.

---

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -652,16 +652,11 @@ fn basic_packing<'a>(
         return Ok(r);
     }
 
-    let mut components: Vec<_> = components.iter().collect();
     let mut r = Vec::new();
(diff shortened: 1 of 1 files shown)


RishabhSaini (approved):
LGTM!

PR #490, opened 2023-05-18
container: Make ManifestDiff have only one lifetime

cgwalters:
This is prep for further changes where I want to compute more derived data.  There's no compelling reason to support disjoint lifetimes here; in the end we need to be OK with the shorter of the two.


RishabhSaini (approved):
LGTM!

PR #491, opened 2023-05-19
Release 0.11.0

cgwalters:
(We should have bumped the semver in git before now)

But I think we're done with API changes for now, so let's go ahead and do the first 0.11 release.


RishabhSaini (approved):
LGTM!

PR #492, opened 2023-06-05
store: Pulling from container-storage needs root privileges

RishabhSaini:
Fixes: https://github.com/coreos/rpm-ostree/issues/4283

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -441,8 +441,13 @@ impl ImageImporter {
         imgref: &OstreeImageReference,
         mut config: ImageProxyConfig,
     ) -> Result<Self> {
-        // Apply our defaults to the proxy config
-        merge_default_container_proxy_opts(&mut config)?;
+        if imgref.imgref.transport == Transport::ContainerStorage {
+            // Fetching from containers-storage, may require privileges to read files
+            merge_default_container_proxy_opts_with_isolation(&mut config, None)?;
+        } else {
+            // Apply our defaults to the proxy config
+            merge_default_container_proxy_opts(&mut config)?;
+        }
         let proxy = ImageProxy::new_with_config(config).await?;
 
         system_repo_journal_print(


RishabhSaini:
Do we need to supply path for [`auth file`](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/src/container/mod.rs#L352) if we already know the source is `container-storage`?
Cane we just completely ignore applying [default configuration to config file](https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/src/container/mod.rs#L334)?

lib/src/container/store.rs
```diff
@@ -441,8 +441,11 @@ impl ImageImporter {
         imgref: &OstreeImageReference,
         mut config: ImageProxyConfig,
     ) -> Result<Self> {
-        // Apply our defaults to the proxy config
-        merge_default_container_proxy_opts(&mut config)?;
+        // Pulling from container-storage doesnt need to drop privileges
+        if imgref.imgref.transport != Transport::ContainerStorage {
+            // Apply our defaults to the proxy config
+            merge_default_container_proxy_opts(&mut config)?;
```
cgwalters, line 447: Hmm.  I think we may instead want to call `merge_default_container_proxy_opts_with_isolation(&mut config, None)` ...the difference here boils down to the authfile handling, which is also irrelevant when pulling from containers-storage...but in the future someone might add *other* config handling there that might be relevant.

cgwalters: IOW more like:

```
        if imgref.imgref.transport != Transport::ContainerStorage {
            // Apply our defaults to the proxy config
            merge_default_container_proxy_opts(&mut config)?;
        } else {
            // When fetching from containers-storage, we can't drop privileges since
            // we may need them to read files.
            merge_default_container_proxy_opts_with_isolation(&mut config, None)?;
        }
```

or so?

lib/src/container/mod.rs
```diff
@@ -333,10 +333,15 @@ impl<'a> ManifestDiff<'a> {
 /// as root, then a default isolation of running the process via `nobody` will be applied.
 pub fn merge_default_container_proxy_opts(
     config: &mut containers_image_proxy::ImageProxyConfig,
+    transport: Transport,
```
cgwalters, line 336: This API is public though, so this change would be a semver break; not sure we want to do another semver break cycle right now.

We could add a new public API though.

(Also then we'll need to change rpm-ostree to call it)

lib/src/container/mod.rs
```diff
@@ -333,10 +333,15 @@ impl<'a> ManifestDiff<'a> {
...
 ) -> Result<()> {
-    let user = rustix::process::getuid()
+    let mut user = rustix::process::getuid()
         .is_root()
```
cgwalters, line 339: I think this would be a bit cleaner as
```
let user = (rustix::process::getuid().is_root && transport != Transport::ContainerStorage).then_some(isolation::DEFAULT_UNPRIVILEGED_USER);
``` 

or so?

cgwalters (changes requested):

lib/src/container/store.rs
RishabhSaini: Agree that this is a cleaner approach

lib/src/container/mod.rs
RishabhSaini: Took the previous approach

lib/src/container/mod.rs
RishabhSaini: took the previous approach

lib/src/container/store.rs
```diff
@@ -441,8 +441,13 @@ impl ImageImporter {
         imgref: &OstreeImageReference,
         mut config: ImageProxyConfig,
     ) -> Result<Self> {
-        // Apply our defaults to the proxy config
-        merge_default_container_proxy_opts(&mut config)?;
+        if imgref.imgref.transport == Transport::ContainerStorage {
+            //Fetching from containers-storage, may require priviliges to read files
```
cgwalters, line 445: Still need space between comment and text here.

cgwalters (changes requested):

lib/src/container/store.rs
RishabhSaini: ah yes! :)

lib/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.10.7"
+version = "0.11.0"
```
cgwalters, line 9: I think you should `git fetch origin && git rebase origin/main` since this commit is already in main.

cgwalters (approved):

PR #493, opened 2023-06-10
container/store: Make baseimage ref prefix `pub`

cgwalters:
Since it's intended for use by external consumers.


jmarrero (approved):
/lgtm

PR #497, opened 2023-07-06
container: Add an API to serialize `Transport`

cgwalters:
This will fix the need demonstrated by
https://github.com/containers/bootc/pull/106/files#diff-c163debeb7fdf9b238e5c589e0000c80684ca9a09abd408242df4bad489626a6R38

We made a historical mistake here in having our `Display` include the `:`...but it's hard to deal with because inherently the format is messy because of the `://` in `docker://`.


jmarrero (approved):
cool

PR #498, opened 2023-07-06
deploy: Add an API to prune undeployed images

cgwalters:
This is part of fixing https://github.com/coreos/rpm-ostree/issues/4391 but is also in the general theme of making things less "stateful".

A huge huge mess today is `rpm-ostree rebase` and `bootc switch` both have `--retain` options which keep the previous image.

But really what we want is to use the deployments as source-of-truth; that way if e.g. an admin pins a deployment, it automatically pins the image too.

And this will help strongly align with the bootc direction in reconciling to desired state.


jmarrero (approved):
lgtm

PR #499, opened 2023-07-11
Add previous container image state

lukewarmtemp:
Prerequisite for the `rpm-ostree upgrade --check` issue: https://github.com/coreos/rpm-ostree/pull/4486

Implemented:
- previous_state `LayeredImageState` object added to the `PreparedImport` object
- `export_as_string()` function is added to the `ManifestDiff` object

Note: The `previous_state` member of the struct is not actually used in the rpm-ostree issue stated above. Only the `export_as_string()` is used. However, the changes may see use in the future and can probably be left in.

See commit message for more details.

### DIFF
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -265,6 +265,18 @@ pub struct ManifestDiff<'a> {
     pub removed: Vec<&'a oci_spec::image::Descriptor>,
     /// Layers which are present in the new image but not the old image.
     pub added: Vec<&'a oci_spec::image::Descriptor>,
+    /// Total number of packages
+    pub total: u64,
+    /// Size of total number of packages.
+    pub total_size: u64,
+    /// Number of packages removed
+    pub n_removed: u64,
+    /// Size of the number of packages removed
+    pub removed_size: u64,
+    /// Number of packages added
+    pub n_added: u64,
+    /// Size of the number of packages added
+    pub added_size: u64,
 }
 
 impl<'a> ManifestDiff<'a> {
@@ -297,11 +309,27 @@ impl<'a> ManifestDiff<'a> {
             }
         }
         added.sort_by(|a, b| a.digest().cmp(b.digest()));
+
+        fn layersum<'a, I: Iterator<Item = &'a oci_spec::image::Descriptor>>(layers: I) -> u64 {
+            layers.map(|layer| layer.size() as u64).sum()
+        }
+        let total = dest_layers.len() as u64;
+        let total_size = layersum(dest.layers().iter());
+        let n_removed = removed.len() as u64;
+        let n_added = added.len() as u64;
+        let removed_size = layersum(removed.iter().copied());
+        let added_size = layersum(added.iter().copied());
         ManifestDiff {
             from: src,
             to: dest,
             removed,
             added,
+            total,
+            total_size,
+            n_removed,
+            removed_size,
+            n_added,
+            added_size,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


lib/src/container/mod.rs
```diff
@@ -324,6 +324,23 @@ impl<'a> ManifestDiff<'a> {
         println!("Removed layers:   {n_removed:<4}  Size: {removed_size_str}");
         println!("Added layers:     {n_added:<4}  Size: {added_size_str}");
     }
+
+    /// Exports the total, removed and added content between two OCI images as a string
+    pub fn export_as_string(&self) -> String {
```
cgwalters, line 329: But are we going to be *parsing* this string elsewhere?  If so, I think I'd say instead we should be returning a proper `struct`.

lukewarmtemp: I parse the string using in the following changes to rpm-ostree: https://github.com/coreos/rpm-ostree/blob/7dc4ba82849bccdeef039c6e1a826e6c03bad8ad/src/lib/rpmostree-db.c#L159-L216 (which is from: https://github.com/coreos/rpm-ostree/pull/4486).

Changing this code to return a struct would just require a few changes to the rpm-ostree PR and would be a lot cleaner compare to how I'm parsing the string. However, is there a way to convert a Rust struct to a C struct using CXX bridge?

cgwalters: Yeah, so what you're hitting is generic C++/Rust integration being painful.  Really, *all* cross-language FFI is painful.

On this specific topic, there is https://cxx.rs/shared.html - however, I don't think it can be used for "foreign" types (i.e. structs in this crate), so we'd need to create another copy of the struct in rpm-ostree, and then convert it on the boundary.

I can help you with this in a realtime call if you want.

BTW what may help us here is to change the implementation of `print` in this crate to use the structs - this would avoid duplicating logic more.

cgwalters: Take a look at e.g. https://github.com/coreos/rpm-ostree/blob/d498bac1dd00344b23ccb2aa8d590c34acc691e2/rust/src/lib.rs#L445 as an example and how that's bridged between Rust/C++.

lib/src/container/mod.rs
```diff
@@ -267,6 +267,23 @@ pub struct ManifestDiff<'a> {
...
+#[derive(Debug)]
+pub struct ExportedManifestDiff {
+    /// Total number of packages
+    pub total: String,
+    /// Size of total number of packages.
+    pub total_size: String,
+    /// Number of packages removed
+    pub n_removed: String,
+    /// Size of the number of packages removed
+    pub removed_size: String,
```
cgwalters, line 280: Shouldn't these be integer values e.g. `u64`?

lib/src/container/mod.rs
```diff
@@ -308,21 +325,31 @@ impl<'a> ManifestDiff<'a> {
...
         let n_added = self.added.len();
         let removed_size = layersum(self.removed.iter().copied());
         let removed_size_str = glib::format_size(removed_size);
         let added_size = layersum(self.added.iter().copied());
         let added_size_str = glib::format_size(added_size);
-        println!("Total new layers: {new_total:<4}  Size: {new_total_size}");
+        println!("Total new layers: {new_total:<4}  Size: {new_total_size_str}");
         println!("Removed layers:   {n_removed:<4}  Size: {removed_size_str}");
         println!("Added layers:     {n_added:<4}  Size: {added_size_str}");
+        let manifest_diff = ExportedManifestDiff {
```
cgwalters, line 344: See https://doc.rust-lang.org/std/keyword.return.html - we don't need explicit `return` in Rust in this situation.  We can just do:

```
ExportedManifestDiff { ... }
``

(But if we follow my first suggestion, we don't need this at all)

File: lib/src/container/mod.rs (hunk 2 above)
Comment (lukewarmtemp) [reply lib/src/container/mod.rs]: I've changed them all the `u64` and made the appropriate changes on the rpm-ostree side.

File: lib/src/container/mod.rs (hunk 3 above)
Comment (lukewarmtemp) [reply lib/src/container/mod.rs]: I assume that "But if we follow my first suggestion, we don't need this at all" refers to moving the members of the `ExportedManifestDiff` directly into the `ManifestDiff` struct?

File: lib/src/container/mod.rs [hunk 4]
```diff
@@ -308,21 +325,31 @@ impl<'a> ManifestDiff<'a> {
 
 impl<'a> ManifestDiff<'a> {
     /// Prints the total, removed and added content between two OCI images
-    pub fn print(&self) {
+    pub fn print(&self) -> ExportedManifestDiff {
```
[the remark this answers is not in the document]
lukewarmtemp: Sounds good. I've implemented this and also made some changes to the print() function to accommodate this change by remove the need to recompute the values already computed in the `ManifestDiff::new` function.

cgwalters (approved):

Issue #163, opened 2021-11-18
test that we support image signatures


cgwalters:
xref https://github.com/containers/skopeo/issues/1482

We should validate that we're doing image signatures via the proxy correctly.

cgwalters:
In this issue, the great thing about the new ostree-native-container flow is that if you have a setup to sign container images, that *exact same setup* can be used to sign OS updates.

See https://docs.podman.io/en/latest/markdown/podman-image-sign.1.html and https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/signing_container_images for some old-style GPG signatures.  As of recently the containers/image stack gained support for "cosign", see https://github.com/containers/skopeo/pull/1701

RishabhSaini:
To test the [recent changes](https://github.com/containers/skopeo/pull/2029) for policy verification:

1) Modify `/etc/containers/policy.json` for `sigstore`/`gpg` signed images from the remote-registry
2) Sign an existing `fcos` image and push to the remore-registry
3) Try doing an `rpm-ostree rebase ${signed-image}`
4) Ensure it fails

Since we do not currently sign `fcos` or any `ostree-based-images`, we need to have signed images available. 

Initially it was thought this can be done locally by doing `skopeo copy docker://quay.io/fedora/fedora-coreos:testing-devel oci:/var/lib/containers/signed-local-registry/sigstore/test.oci --sign-by-sigstore-private-key fcos.key
`. 
But this fails unfortunately, since how sigstore signs images is by pushing the artifacts generated to the remote-registry. Hence, signing an local `oci` or `dir` does not work. It instead gives the following error: `Cannot determine canonical Docker reference for destination oci:/var/lib/containers/signed-local-registry/sigstore/test.oci`. 

Instead we need to be able to push this to some ephemeral testing Docker image registry. The perfect candidate was ttl.sh as mentioned in sigstore doumentation, but unfortunately `fcos` image exceeds the maxiumum image size limit there. Is there any other registry we could push to and verify instead?

cgwalters:
So CI on this repository mainly uses GHA, for which there is https://docs.github.com/en/actions/using-containerized-services/about-service-containers

But that's just sugar for running a container...we can run any registry (quay.io, docker/distribution or whatever) inside a GHA job right?

PR #500, opened 2023-07-17
Add test to ensure policy verification occurs when pulling image

RishabhSaini:
Fixes: #163

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
 derived_img=oci:/var/tmp/derived.oci
 systemd-run -dP --wait skopeo copy containers-storage:localhost/fcos-derived "${derived_img}"
 
+# Verify policy
+cat > /etc/containers/registries.d/default.yaml << EOF
+docker:
+  quay.io/rh_ee_rsaini/coreos:
+    use-sigstore-attachments: true
+EOF
+
+cat > /etc/containers/policy.json << EOF
+{
+    "default": [
+        {
+            "type": "reject"
+        }
+    ],
+    "transports": {
+        "docker": {
+            "quay.io/fedora/fedora-coreos": [
+                {
+                    "type": "insecureAcceptAnything"
+                }
+            ],
+            "quay.io/rh_ee_rsaini/coreos": [
+                {
+                    "type": "sigstoreSigned",
+                    "keyPath": "/etc/pki/containers/fcos.pub",
+                    "signedIdentity": {
+                        "type": "matchRepository"
+                    }
+                }
+            ]
+
+        }
+    }
+}
+EOF
+mkdir repo
+ostree --repo=repo init --mode=bare
+mkdir -p /etc/pki/containers
+#Ensure Wrong Public Key fails
+cat > /etc/pki/containers/fcos.pub << EOF
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEPw/TzXY5FQ00LT2orloOuAbqoOKv
+relAN0my/O8tziGvc16PtEhF6A7Eun0/9//AMRZ8BwLn2cORZiQsGd5adA==
+-----END PUBLIC KEY-----
+EOF
+ostree container image pull repo ostree-image-signed:docker://quay.io/rh_ee_rsaini/coreos 2> error
+test "$(grep -c "invalid signature" error)" -eq 1
+
+#Ensure Correct Public Key succeeds
+cat > /etc/pki/containers/fcos.pub << EOF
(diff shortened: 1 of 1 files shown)


ci/priv-integration.sh
```diff
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
...
+        "docker": {
+            "quay.io/fedora/fedora-coreos": [
+                {
+                    "type": "insecureAcceptAnything"
+                }
+            ],
+            "quay.io/rh_ee_rsaini/coreos": [
+                {
+                    "type": "sigstoreSigned",
+                    "keyPath": "/etc/pki/containers/fcos.pub",
```
cgwalters, line 112: While this doesn't matter functionally, I'd write the key *first* before writing this policy file.

ci/priv-integration.sh
```diff
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
...
+                        "type": "matchRepository"
+                    }
+                }
+            ]
+
+        }
+    }
+}
+EOF
+mkdir repo
```
cgwalters, line 123: Hmm, any reason not to use `/sysroot/ostree/repo` as we're doing in other cases?  (This is indirectly happening via `$sysroot`).  A small advantage of this is it'd share any common layers.

ci/priv-integration.sh
```diff
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
...
+mkdir -p /etc/pki/containers
+#Ensure Wrong Public Key fails
+cat > /etc/pki/containers/fcos.pub << EOF
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEPw/TzXY5FQ00LT2orloOuAbqoOKv
+relAN0my/O8tziGvc16PtEhF6A7Eun0/9//AMRZ8BwLn2cORZiQsGd5adA==
+-----END PUBLIC KEY-----
+EOF
+ostree container image pull repo ostree-image-signed:docker://quay.io/rh_ee_rsaini/coreos 2> error
+test "$(grep -c "invalid signature" error)" -eq 1
```
cgwalters, line 134: Hmm but doesn't the process exit with an error code?  Doesn't this need to be more like:
```
if ostree container image pull repo ostree-image-signed:docker://quay.io/rh_ee_rsaini/coreos 2> error; then
  echo "unexpectedly pulled image" 1>&2
  exit 1
else
  test "$(grep -c "invalid signature" error)" -eq 1
fi
```

cgwalters: We could also sanity check that `ostree container image list` doesn't have the image.

cgwalters: Finally, I don't understand why we are using `grep -c` here instead of just e.g.:

`grep -q "invalid signature" error`

?

ci/priv-integration.sh
```diff
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
...
+
+#Ensure Correct Public Key succeeds
+cat > /etc/pki/containers/fcos.pub << EOF
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEREpVb8t/Rp/78fawILAodC6EXGCG
+rWNjJoPo7J99cBu5Ui4oCKD+hAHagop7GTi/G3UBP/dtduy2BVdICuBETQ==
+-----END PUBLIC KEY-----
+EOF
+ostree container image pull repo ostree-image-signed:docker://quay.io/rh_ee_rsaini/coreos > out
+test "$(grep -c "Wrote: ostree-image-signed:docker://quay.io/rh_ee_rsaini/coreos" out)" -eq 1
```
cgwalters, line 144: I think instead of parsing the stdout we should verify via `ostree container image history` e.g.

ci/priv-integration.sh
```diff
@@ -85,6 +85,64 @@ systemd-run -dP --wait podman build -t localhost/fcos-derived .
 derived_img=oci:/var/tmp/derived.oci
 systemd-run -dP --wait skopeo copy containers-storage:localhost/fcos-derived "${derived_img}"
 
+# Verify policy
+cat > /etc/containers/registries.d/default.yaml << EOF
+docker:
+  quay.io/rh_ee_rsaini/coreos:
```
cgwalters, line 91: Probably worth having this up at the top of the file as a constant like 
```
# This image was generated manually; TODO auto-generate in quay.io/coreos-assembler or better start sigstore signing our production images
FIXTURE_SIGSTORE_SIGNED_FCOS_IMAGE=quay.io/rh_ee_rsaini/coreos
```

cgwalters (approved):

Issue #502, opened 2023-07-19
deploy: Inherit previous kargs by default


cgwalters:
Today `ostree container image deploy` doesn't preserve previous deployment kargs, it only preserves booted: https://github.com/ostreedev/ostree-rs-ext/blob/c2a292cfb5b6c9c8c0f20ad1d457d4655221ed79/lib/src/container/deploy.rs#L86

This is a huge ergonomic problem for https://github.com/coreos/fedora-coreos-tracker/issues/1151#issuecomment-1430263412

PR #503, opened 2023-07-19
deploy: Don't overwrite kargs by default

cgwalters:
This is the same bug as
https://github.com/ostreedev/ostree-rs-ext/commit/3089166a4456cdcfa0568aedcda31eac65ac00ee but for the not-booted case.

Basically in the C API bridged to Rust we can't distinguish between "NULL array" and "zero length array".  But the _with_options path supports distinguishing them, and we want the "no kargs provided" case to not override anything.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/502


cgwalters:
(Not tested yet)

jmarrero (approved):
lgtm

cgwalters:
Related to https://github.com/coreos/fedora-coreos-tracker/issues/1151#issuecomment-1662203876

PR #504, opened 2023-07-19
repair: New functionality to detect (future: fix) inodes

cgwalters:
Initial code to detect the situation resulting from https://github.com/ostreedev/ostree/pull/2874/commits/de6fddc6adee09a93901243dc7074090828a1912

EDIT: Here's instructions to run this:

> podman run --rm -ti --pull=newer --privileged -v /:/rootfs --net=none quay.io/cgwalters/ostree-ext-dev provisional-repair repair --sysroot /rootfs/sysroot --dry-run

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -8,10 +8,12 @@
 use anyhow::{Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::{Parser, Subcommand};
+use fn_error_context::context;
 use ostree::{cap_std, gio, glib};
 use std::collections::BTreeMap;
 use std::ffi::OsString;
 use std::path::PathBuf;
+use std::process::Command;
 use tokio::sync::mpsc::Receiver;
 
 use crate::commit::container_commit;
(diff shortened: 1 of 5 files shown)


cgwalters:
My plan is to build this code into a container image and ship it alongside a fixed ostree to repair.

Brief notes on running this so far outside of a container image

```
$ unshare -m  # create new mountns
$ mount -o remount,rw /sysroot
$ ostree-ext-cli provisional-repair --sysroot / --dry-run
```

The happy path results in:
```
Attempting analysis of ostree state for files that may be incorrectly linked
For more information, see https://github.com/ostreedev/ostree/pull/2874/commits/de6fddc6adee09a93901243dc7074090828a1912

Gathering inodes...
No colliding objects found.
```

But when I boot e.g. FCOS with a large filesystem *and* rebase to a container image, I get warnings (as expected).

cgwalters:
(Let's keep this as draft for now, don't need to ship it in git main yet)

jmarrero (approved):
lgtm

cgwalters:
OK what I struggled with a bit is that at a technical level in order to use this code we need to make it a `pub`lic API so we could ship it in rpm-ostree, but we can cut that gordian knot by having it be `pub`, but also just `#[doc(hidden)]` and we can not consider it a semver break if we change/remove it.

I think longer term it'd make sense to split this type of code into a separate "fsck" like tool, but as of right now it kind of pokes into the internals of things (as it needs to today).

Anyways I lifted draft and let's go ahead and merge this.

jmarrero (approved):

cgwalters:
Just to link things, https://github.com/ostreedev/ostree-rs-ext/pull/505 is continuation of this work.

aospan:
I swiftly reviewed the changes made in this commit. Excellent job! It's relieving to see that it's possible to fix ostree without having to nuke everything :)

PR #505, opened 2023-07-21
repair: Rework to be more introspectable

cgwalters:
I plan to "productize" this repair code a bit more in OpenShift at least, and I think other admins may want to do similar outside of that too.

In order to make that more reliable:

- Better split the "fsck/--dry-run" path from "repair" i.e mutation
- Introduce a `--write-result-to` argument that outputs JSON.  This allows us to better distinguish the tristate of "OK" from "corruption detected" to "tool failed for some other reason"

Further:

- Drop the check for derived commits, let's just *always* check the inodes because it's not very expensive in the end and it's just really useful to do.
- Add checks for whether the booted deployment is potentially affected, as this is important information; we'll need to redeploy and reboot if so

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -12,6 +12,7 @@ use fn_error_context::context;
 use ostree::{cap_std, gio, glib};
 use std::collections::BTreeMap;
 use std::ffi::OsString;
+use std::io::BufWriter;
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

PR #513, opened 2023-08-14
container: Make requiring bootable field be opt-in

cgwalters:
This fixes a regression introduced by
https://github.com/ostreedev/ostree-rs-ext/commit/9c4a75b3778a3f2fdece095f8f5f7a6289ab512dLooks

Brought up in
https://github.com/ostreedev/ostree/discussions/2978

Basically we need to make this opt-in at higher levels because encapsulating a non-bootable commit (as well as a commit that historically doesn't have that label) must be supported.

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -148,6 +148,8 @@ pub struct ImageImporter {
     target_imgref: Option<OstreeImageReference>,
     no_imgref: bool,  // If true, do not write final image ref
     disable_gc: bool, // If true, don't prune unused image layers
+    /// If true, require the image has the bootable flag
+    require_bootable: bool,
     pub(crate) proxy_img: OpenedImage,
 
     layer_progress: Option<Sender<ImportProgress>>,
@@ -349,11 +351,6 @@ pub(crate) fn parse_manifest_layout<'a>(
     config: &ImageConfiguration,
 ) -> Result<(&'a Descriptor, Vec<&'a Descriptor>, Vec<&'a Descriptor>)> {
     let config_labels = super::labels_of(config);
-    let bootable_key = *ostree::METADATA_KEY_BOOTABLE;
-    let bootable = config_labels.map_or(false, |l| l.contains_key(bootable_key));
-    if !bootable {
-        anyhow::bail!("Target image does not have {bootable_key} label");
-    }
 
     let first_layer = manifest
         .layers()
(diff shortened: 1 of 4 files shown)


lib/src/container/store.rs
```diff
@@ -349,11 +351,6 @@ pub(crate) fn parse_manifest_layout<'a>(
     config: &ImageConfiguration,
 ) -> Result<(&'a Descriptor, Vec<&'a Descriptor>, Vec<&'a Descriptor>)> {
     let config_labels = super::labels_of(config);
-    let bootable_key = *ostree::METADATA_KEY_BOOTABLE;
-    let bootable = config_labels.map_or(false, |l| l.contains_key(bootable_key));
-    if !bootable {
```
omertuc, line 354: The motivation for 9c4a75b3778a3f2fdece095f8f5f7a6289ab512d said ```We want to give a clear and useful error when someone tries
to pull a non-ostree-based container image.```, which is something we no longer have after this code is removed. Bootable might have been too strict, but would we still want some way to check whether it's an ostree image at all? Maybe we should at-least check that the `ostree.commit`  is non-empty

cgwalters: The checking still exists - it wasn't removed.  It just becomes opt in.  The deploy code here is changed to opt in.  And I'll make the corresponding change in bootc and rpm-ostree to do so as well there.

omertuc: I mean even when not deploying, when we're doing a pull with `container unencapsulate`, we would probably want to verify that the image is an actual ostree image and not some other random OCI image

cgwalters: Yes, this is a good thing to discuss.  What I think happened here is not long after the original commit we landed https://github.com/ostreedev/ostree-rs-ext/pull/424/commits/a63171ccd63de0086552cfffc8764d0066023a7f
which as a side effect introduced a hard requirement on the `ostree.encapsulated` label.

With the code built in this PR:

```
$ ./target/release/ostree-ext-cli container unencapsulate --repo=/var/tmp/testrepo ostree-unverified-registry:docker.io/busybox
   Downloading...                                                                                                                                                                                                          error: Fetching manifest: Parsing manifest layout: No ostree.final-diffid label found, not an ostree-bootable container
```

jmarrero (approved):
/lgtm

PR #514, opened 2023-08-14
store: Tweak error message for non-ostree containers

cgwalters:
Since in theory we support non-bootable cases, and we actually have a distinct label for the bootable one.


jmarrero (approved):
lgtm

PR #515, opened 2023-08-16
lib: Dispatch to monomorphic function

cgwalters:
I was looking at `cargo bloat` and this showed up near the top twice. It's a large function and only the option parsing needs to be generic.


cgwalters:
@jmarrero We were talking about binary size of rpm-ostree, this is just one tiny dent in that.

PR #516, opened 2023-08-21
cli: Make `--sysroot` optional

cgwalters:
To make it less likely people run into the footgun of specifying `--sysroot /sysroot` which is buggy right now
https://github.com/ostreedev/ostree/issues/2992


jmarrero (approved):
lgtm

PR #519, opened 2023-08-28
cli/unencapsulate: Add proxy options

cgwalters:
We need to support all the proxy options so that custom authentication files can be supported, etc.

Closes: https://github.com/ostreedev/ostree/issues/3015

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -88,6 +88,9 @@ pub(crate) enum ContainerOpts {
         #[clap(long, value_parser)]
         repo: Utf8PathBuf,
 
+        #[clap(flatten)]
+        proxyopts: ContainerProxyOpts,
+
         /// Image reference, e.g. registry:quay.io/exampleos/exampleos:latest
         #[clap(value_parser = parse_imgref)]
         imgref: OstreeImageReference,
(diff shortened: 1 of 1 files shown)


cgwalters:
```
    Compiling containers-image-proxy v0.5.3
error[E0716]: temporary value dropped while borrowed
   --> /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/containers-image-proxy-0.5.3/src/imageproxy.rs:317:22
    |
317 |                 &mut [iov],
    |                      ^^^^^ creates a temporary value which is freed while still in use
...
320 |             )?;
    |               - temporary value is freed at the end of this statement
321 |             // SAFETY: We provided a buffer
322 |             let iov = r.iovs().next().unwrap();
    |                       -------- borrow later used here
    |
help: consider using a `let` binding to create a longer lived value
    |
315 ~             let binding = [iov];
316 ~             let r = nixsocket::recvmsg::<()>(
317 |                 sockfd.as_raw_fd(),
318 ~                 &mut binding,
    |

For more information about this error, try `rustc --explain E0716`.
error: could not compile `containers-image-proxy` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

Er...this just suddenly popped up; the relevant code here hasn't changed in a while.  Presumably this is something with a new Rust toolchain since 1.72 just came out, but I'm not immediately reproducing it with that version here.  In fact, even if I try to pull the current fcos buildroot container, it doesn't reproduce...weird.

jmarrero (approved):
/lgtm

PR #520, opened 2023-08-28
Cargo.toml: Set `resolver = 2`

cgwalters:
There's a warning in newer Rust 1.72 about this, we do want the v2
resolver.


cgwalters:
oooo our own CI was actually relying on the old feature unification.  Horray for CI though

jmarrero (approved):
lgtm

PR #521, opened 2023-08-28
Bump rust-version = 1.70 && ci: Drop hard errors on clippy warnings by default

cgwalters:
Bump rust-version = 1.70

clap_builder now pulls it in, and we need to follow that train...

---

ci: Drop hard errors on clippy warnings by default

It's just too much churn for too little value.

---

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,8 +15,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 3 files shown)


cgwalters:
Holy cow the chain of CI/build fallout is insane.  Next up is https://github.com/coreos/fedora-coreos-config/pull/2568
and for now I commented it out

jmarrero (approved):
lgtm

PR #522, opened 2023-08-28
chunking: Add const for minimum, change to regular error

cgwalters:
Avoid a panic in this case as it can be reachable via external input.

xref https://github.com/coreos/rpm-ostree/issues/4530


jmarrero (approved):
lgtm

PR #523, opened 2023-08-28
cli/encapsulate: Add `--authfile`

cgwalters:
Ideally we'd pass through all of the proxy options here, but doing that sanely really requires being able to do *pushes* through containers-image-proxy-rs, which is a quite nontrivial amount of work.

For now, let's pass through `--authfile` which is the main thing people want.  Anything else can be worked around by encapsulating to `oci` and then doing a `skopeo copy` from there.

cc https://github.com/ostreedev/ostree/issues/3015


jmarrero (approved):
lgtm

PR #526, opened 2023-08-31
sysroot: Add `from_assume_locked`

cgwalters:
There's a mess with `SysrootLock` because in practice, way, way too many APIs just take the plain old C `Sysroot` object.

In this project, we added a new API that requires a `SysrootLock`, which works well when everything is using ostree-ext from the start, as it is in bootc.

However in rpm-ostree we acquire the lock from C code, but want to call `remove_undeployed_images` which wants `SysrootLock`.  The only practical way out of this is to add an API which asserts that the sysroot is locked and returns this wrapper.

What would actually work better here is to drive this locking logic down into the C library.


jmarrero (approved):
lgtm

PR #529, opened 2023-08-31
Use cap_tempfile via cap_std_ext

cgwalters:
This way things like Dependabot understand there's only one dependency to bump instead of multiple that must move in lockstep.


jmarrero (approved):
lgtm

PR #531, opened 2023-09-08
chunking: Ignore ostree commit layer

cgwalters:
I was hitting a panic in this section of the code; I think
triggering it involves removing packages, but I haven't narrowed
it down more precisely.

Hopefully at some point I will try to add some more unit
testing for this...

---


jmarrero (approved):
lgtm

PR #532, opened 2023-09-08
lib: Bump to ostree 0.19, cap-std-ext 3.0 && lib: Bump to rustix 0.38 && Bump semver to 0.12

cgwalters:
keyfileext: Drop optional_string_list

Nothing is using this, and the underlying glib type changed
so it's annoying to keep.

---

lib: Bump to ostree 0.19, cap-std-ext 3.0

This is just "keeping up with churn":

- Newer versions of glib crates
- Bump to cap-std-ext 0.3, which bumps cap-std 2.0

---

lib: Bump to rustix 0.38

Again just keeping up.

---

lib: Bump semver to 0.12

Because we bumped ostree.

---

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -34,7 +34,7 @@ openssl = "0.10.33"
 ostree = { features = ["v2022_6"], version = "0.19.0" }
 pin-project = "1.0"
 regex = "1.5.4"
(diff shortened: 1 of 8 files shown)


jmarrero (approved):
lgtm

PR #533, opened 2023-09-09
container: Canonicalize imgref to ostree-unverified-registry

cgwalters:
xref https://github.com/openshift/machine-config-operator/pull/3857#issuecomment-1712026017

The MCO has code which incorrectly parses these image references; because rpm-ostree defaults to writing `ostree-unverified-registry`, let's canonicalize to that because it looks better too.


jmarrero (approved):
lgtm

PR #534, opened 2023-09-10
container: Move more metadata into `ExportOpts`

cgwalters:
container: Move more metadata into `ExportOpts`

This drops two extra arguments we added over time; in
a few places before we had e.g. `None, None, None` being passed
which just looks awkward.  And we also threaded through all 3
in various places.

The `ExportOpts` just needs to grow a lifetime argument, but that
turned out to not be too bad when I realized we could use the
elided lifetime `<'_>` in all methods that use it.

---

container: Make `ExportOpts` `#[non_exhaustive]`

It's intended for this purpose, where we want to be able
to add elements to the struct over time without breaking semver.

---

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -269,7 +269,7 @@ impl Chunking {
     pub fn from_mapping(
         repo: &ostree::Repo,
         rev: &str,
-        meta: ObjectMetaSized,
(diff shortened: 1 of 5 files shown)


jmarrero (approved):
lgtm

PR #536, opened 2023-09-11
deploy: Use booted stateroot if available

cgwalters:
We really want the semantics here to more closely match `bootc switch` and `rpm-ostree rebase` etc.  This is just the "offline capable" version.  But if run from a booted system let's DTRT.


jmarrero (approved):
lgtm

Issue #496, opened 2023-07-06
change importer to support caching new manifests/configs without layers


cgwalters:
Today in the underlying ostree code, there's the concept of a ["partial" commit](https://docs.rs/ostree/latest/ostree/struct.RepoCommitState.html#associatedconstant.PARTIAL).  This happens when we fetch the metadata and are still in progress downloading the data.

This concept nicely maps to "check for updates" - one can download just the new commit object and persistently cache the queued update.

We don't have a similar concept in the ostree-container layer.  Today `ImageImporter::prepare` will fetch the new manifest+config, but it's only stored in transient memory by default.

Instead, perhaps we could add a new API like `prepare_and_persist` that updates the commit object only and also flags it as partial explicitly.  We'd also want to propagate the concept of a "partial" image into higher level APIs such as `LayeredImageState` so it can be rendered by clients.

cgwalters:
WIP for this

```
diff --git a/lib/src/container/store.rs b/lib/src/container/store.rs
index fd66359e..9abd04cb 100644
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -20,6 +20,7 @@ use oci_spec::image::{self as oci_image, Descriptor, History, ImageConfiguration
 use ostree::prelude::{Cast, FileEnumeratorExt, FileExt, ToVariant};
 use ostree::{gio, glib};
 use rustix::fs::MetadataExt;
+use serde::{Deserialize, Serialize};
 use std::collections::{BTreeSet, HashMap};
 use std::iter::FromIterator;
 use tokio::sync::mpsc::{Receiver, Sender};
@@ -188,6 +189,17 @@ impl ManifestLayerState {
     }
 }
 
+/// Information about an already fetched cached new manifest and config.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CachedPreparedImport<'a> {
+    /// The manifest digest that was found
+    pub manifest_digest: Cow<'a, str>,
+    /// The deserialized manifest.
+    pub manifest: Cow<'a, oci_image::ImageManifest>,
+    /// The deserialized configuration.
+    pub config: Cow<'a, oci_image::ImageConfiguration>,
+}
+
 /// Information about which layers need to be downloaded.
 #[derive(Debug)]
 pub struct PreparedImport {
@@ -437,6 +449,11 @@ fn timestamp_of_manifest_or_config(
 }
 
 impl ImageImporter {
[...]

PR #537, opened 2023-09-13
container: Cache new manifest/config in prepare, add API to query

cgwalters:
Closes: https://github.com/ostreedev/ostree-rs-ext/issues/496

In https://github.com/coreos/rpm-ostree/pull/4486 we were working on fixing `rpm-ostree upgrade --check` with containers.

However, what we really want here is to *persist* the updated manifest (and config) that we fetch.  And if we do that, we might as well just make it part of the current `prepare()` API so it happens automatically.

In this change, we do so via detached commit metadata.  An important thing here is that the data is then automatically lifecycle bound to the merge commit - and the merge commit always changes when we fetch a new manifest.

In order to do an offline query (e.g. in rpm-ostree we want to re-synthesize a higher level summary of the queued update) add an API which allows querying a previously saved cached update.

Hence a flow like this should work:

- OS boots
- OS updater does a background "check for updates" via calling `prepare()`
- OS updater finds an update, and renders metadata to the user or orchestration system
- <time passes; OS update is not downloaded - e.g. user is on metered data or whatever>
- system reboots for other reasons
- OS updater can re-render the fact that a queued update was found *without* touching the network
- User can initiate a full fetch (e.g. including image layers) targeting *exactly* the previously prepared fetch.  This makes things much more race-free; if the image was GC'd in the meantime we correctly fail.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -7,7 +7,9 @@
 
 use anyhow::{Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
+use cap_std::fs::Dir;
 use cap_std_ext::cap_std;
+use cap_std_ext::prelude::CapStdExtDirExt;
 use clap::{Parser, Subcommand};
 use fn_error_context::context;
 use io_lifetimes::AsFd;
@@ -220,6 +222,12 @@ pub(crate) enum ContainerImageOpts {
         /// Don't display progress
         #[clap(long)]
         quiet: bool,
+
+        /// Just check for an updated manifest, but do not download associated container layers.
+        /// If an updated manifest is found, a file at the provided path will be created and contain
+        /// the new manifest.
+        #[clap(long)]
+        check: Option<Utf8PathBuf>,
     },
 
     /// Output metadata about an already stored container image.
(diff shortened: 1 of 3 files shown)


cgwalters:
Thinking about this a bit more, two potential issues:

- We don't offer a direct API to clear cached updates...not sure anyone will care, but I could imagine it
- This actually breaks the use case for just checking for updates from a *read-only* repository, but again, it's not really clear that's a real use case

jmarrero:
Why not being able to check for updates from a read-only repo is an issue? 
I assume once you rebase to it, you would need to rebase to another one to get new content right? (I am guessing that read-only means the contents of the repo will not change.)

cgwalters:
The "read only repo" case happens to day on e.g. the FCOS Live ISO.  You can do `rpm-ostree status`, but not `rpm-ostree install` e.g.  Now...eventually we probably *should* make the repo (and everything else) transiently writable with an overlayfs just like `etc` but that's a separate ball of wax.

OK well, this doesn't work today (try with e.g. `cosa run --qemu-iso fedora-coreos-latest-live.x86_64.iso`):

```
$ rpm-ostree upgrade --check
error: Remounting /sysroot read-write: Permission denied
```

And so this won't be a regression.

jmarrero:
As for _"We don't offer a direct API to clear cached updates..."_
If people can rm the files if needed, this lgtm.

jmarrero (approved):

cgwalters:
Yeah, one can use the low level ostree API to clear detached metadata entirely at least.

PR #538, opened 2023-09-15
two minor patches

cgwalters:
lib: Add a few semver-break cookies

So we know what to change for the next semver break; all minor
cleanups.

---

cli: Fix a doc link

Just noticed this running `cargo doc --lib`.

---


jmarrero (approved):
lgtm

PR #540, opened 2023-09-18
commit: Skip bind mounts too

cgwalters:
This fixes a failure I was seeing using `buildah` in Gitlab CI; it seems like `/run/secrets` ends up as a bind mount here, and because it's a bind mount it has the same device node.

With new enough Linux there's a handy attribute from the `statx` system call that will tell us whether a given path is a mount point root.  Let's use that to start.

(The other alternatives are much uglier; see the linked systemd
 git source code)

We may need to end up doing some of that if we find we keep hitting this with systems that have older kernels.


cgwalters:
cc https://github.com/bytecodealliance/rustix/pull/831

jmarrero (approved):
lgtm

PR #542, opened 2023-09-19
lib: Fix mountpoint check

cgwalters:
I got confused by the statx API here; it just happened to work because we were ignoring the mask.

This version comes with a test case too.


jmarrero (approved):
/lgtm

PR #544, opened 2023-09-25
Allows container deployment from local directory

lukewarmtemp:
This patch allows container image deployment stored in a local directory in addition to the supported remote OCI registry, OCI archive tarball, and local container storage. For example, the following command is now supported:

`sudo ostree container image deploy --imgref ostree-unverified-image:dir:/var/home/core/fcos --sysroot /var/home/core/sysroot --stateroot fedora-coreos`

The reason for making this change relates to the effort to create an FCOS qemu image from a container image input using OSBuild. OSBuild stores container image sources in a local directory in order to solve another bug and, thus, should not be changed. Therefore, to allow OSBuild to deploy from a container image, this patch is needed.

### DIFF
--- a/lib/src/cli.rs
+++ b/lib/src/cli.rs
@@ -320,6 +320,7 @@ pub(crate) enum ContainerImageOpts {
 
         /// Source image reference, e.g. ostree-remote-image:someremote:registry:quay.io/exampleos/exampleos@sha256:abcd...
         /// This conflicts with `--image`.
+        /// This conflicts with `--image`. Supports `registry:`, `docker://`, `oci:`, `oci-archive:`, `containers-storage:`, and `dir:`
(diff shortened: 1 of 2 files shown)


cgwalters:
Minor style note; we tend to use "Linux kernel style" git commits here, which use a "topic 
prefix".  Also the subject should be imperative.

Something like:

> container: Add support for the `dir` transport

Procedurally otherwise, I think this makes sense, with the caveat that 

>        dir:path
>         An  existing local directory path storing the manifest, layer tarballs and signatures as individual files. This is a non-standardized format, primarily useful for debugging or non‐
>       invasive container inspection.

- Please add some basic unit tests for parsing the transport
- It'd also be nice to add an integration test

cgwalters (approved):

PR #546, opened 2023-09-25
container: Add more tests for image reference

cgwalters:
- Cover containers-storage and oci-archive too
- Test that we can reserialize back into the same string


cgwalters:
cc @lukewarmtemp

jmarrero (approved):
lgtm

PR #547, opened 2023-09-25
container: Add dir container deploy test 

lukewarmtemp:
This change adds an integration test for deploying from a container image stored in a local directory. It is directly related a recent change to support the `dir:` transport when using `ostree container image deploy`: https://github.com/ostreedev/ostree-rs-ext/pull/544

### DIFF
--- a/ci/priv-integration.sh
+++ b/ci/priv-integration.sh
@@ -86,7 +86,9 @@ RUN touch /usr/share/somefile
 EOF
 systemd-run -dP --wait podman build -t localhost/fcos-derived .
 derived_img=oci:/var/tmp/derived.oci
+derived_img_dir=dir:/var/tmp/derived.dir
 systemd-run -dP --wait skopeo copy containers-storage:localhost/fcos-derived "${derived_img}"
+systemd-run -dP --wait skopeo copy "${derived_img}" "${derived_img_dir}"
 
 # Prune to reset state
 ostree refs ostree/container/image --delete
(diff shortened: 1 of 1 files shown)


ci/priv-integration.sh
```diff
@@ -106,6 +108,19 @@ img_commit2=$(ostree --repo=${repo} rev-parse ostree/container/image/${imgref})
 test "${img_commit}" = "${img_commit2}"
 echo "ok deploy derived container identical revs"
 
+ostree-ext-cli container image deploy --sysroot "${sysroot}" \
```
cgwalters, line 111: I think it'd be a bit better to do e.g.
```
skopeo copy ${derived_img} ${derived_img_dir}
```

here (*instead* of the above `systemd-run skopeo copy`), and then we can do the deployment...

ci/priv-integration.sh
```diff
@@ -106,6 +108,19 @@ img_commit2=$(ostree --repo=${repo} rev-parse ostree/container/image/${imgref})
...
+        --stateroot "${stateroot}" --imgref ostree-unverified-image:"${derived_img_dir}"
+imgref=$(ostree refs --repo=${repo} ostree/container/image | head -1)
+img_commit=$(ostree --repo=${repo} rev-parse ostree/container/image/${imgref})
+ostree-ext-cli container image remove --repo "${repo}" "${derived_img_dir}"
+
+ostree-ext-cli container image deploy --sysroot "${sysroot}" \
```
cgwalters, line 117: Then here I'd do:
```
rm -rf /var/tmp/derived.dir
```
or so to avoid "leaking" double disk space usage across the rest of the test.

I think we can also drop the re-deploy and test of identical commit ID, because that's very unlikely to break just with the `dir` transport.
.

cgwalters (changes requested):

cgwalters:
BTW one tip to test Github Action changes is to push them to *your fork*'s `main` branch, then you can see the [runs in your fork](https://github.com/lukewarmtemp/ostree-rs-ext/actions) without the noise of pushing to a PR.

cgwalters (approved):

PR #551, opened 2023-10-02
lib: Add mountutil module

cgwalters:
I copy/pasted the mountpoint check into bootc, but it'd be better to centralize.


jmarrero (approved):
lgtm

PR #553, opened 2023-10-03
Quiet extremely verbose tracing::debug

cgwalters:
tree-wide: Clean up usage of `tracing::instrument`

We're really verbose with `env RUST_LOG=debug`...most of
these spans shouldn't be emitting any arguments at all . And
even then let's lower to debug level for them too.

---

store: Silence extremely verbose tracing::debug

This ends up recursing into the manifest/config objects
which is a wall of spam with `env RUST_LOG=debug`.  Just print
the merge commit to start here.

---


jmarrero (approved):
lgtm

PR #555, opened 2023-10-09
chunking: Handle low amount of high-size packages

travier:
If we have a low amount of high-size packages then we will drain with an index higher than the size of the Vec which triggers a panic.

Fixes: https://github.com/coreos/rpm-ostree/issues/4646

### DIFF
--- a/lib/src/chunking.rs
+++ b/lib/src/chunking.rs
@@ -462,8 +462,12 @@ fn get_partitions_with_threshold<'a>(
     }
 
     // Extra high-size packages
-    let mut remaining_pkgs: Vec<_> = high_size.drain(limit_hs_bins..).collect();
-    assert_eq!(high_size.len(), limit_hs_bins);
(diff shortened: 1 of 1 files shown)


travier:
Note: Untested yet.

travier:
```
thread 'chunking::test::test_advanced_packing' panicked at 'assertion failed: high_size.len() < limit_hs_bins', lib/src/chunking.rs:470:5
Found file: var/foo
Found file: var/foo
Found file: var/lib/nested/foo
thread 'chunking::test::test_packing_fcos' panicked at 'assertion failed: high_size.len() < limit_hs_bins', lib/src/chunking.rs:470:5
```

Looks like this assert just found another off by one in my change.

cgwalters (approved):
I think this makes sense...but man, we're going to need a lot more tests here.

travier:
Verified in https://github.com/travier/ostree-containers/pull/1 that this fixes this issue.

PR #556, opened 2023-10-20
deploy: Always use merge commit

cgwalters:
I don't recall now the rationale for the previous code; if we're using this path we always want to be in a container-native mode.  Currently the "unpack to pure ostree commit" gets triggered if one uses this on a container image created via raw `ostree container encapsulate`.

Anyone who wants the prior behavior should instead unencapsulate and deploy the underlying commit.


jlebon (approved):
Seems sane to me.

PR #558, opened 2023-10-28
tar: Fix multiple hardlinks

cgwalters:
The previous commit here (43e1648e97ef82f2cb86fd0813a5f338585eea97) handled the case of *one* hardlink, but the logic for the secondary one was backwards.

I noticed this when generating a derived image, in some cases we only end up with `/usr/lib/sysimage/rpm-ostree-base-db` as modified, which causes really confusing semantics that look like packages going missing.


jmarrero (approved):
lgtm

PR #559, opened 2023-10-29
tar: Fix hang if ostree exits with failure

cgwalters:
A classic bug; we need to concurrently handle both the process exit status as well as output from its stdout/stderr streams.

I hit this hang when running `ostree-ext-cli container pull` directly and the target sysroot mount isn't writable.


jmarrero (approved):
lgtm

PR #565, opened 2023-11-09
unencapsulate: Handle legacy docker MIME type

cgwalters:
In the original design of the skopeo proxy, the idea is that it shields us from the legacy Docker formats, converting everything to OCI.

That has worked - except it turns out in the case of pulling a docker-formatted image from `containers-storage`, the proxy code fails to convert the layer types.

In this case, it's easy to handle here; we only need to care about the uncompressed version in this case.


jmarrero (approved):
lgtm

PR #566, opened 2023-11-10
isolation: Also reset environment

cgwalters:
Otherwise in some cases the containers/image stack can try to look at things like `$HOME` which might be set to `/root` if we're running outside of systemd, and that can cause permission denials.


jmarrero (approved):
lgtm

srd424:
This looks like it broke passing of http_proxy / https_proxy through to skopeo - any ideas of a work-around?

PR #567, opened 2023-11-28
encapsulate: Support setting arbitrary container config

cgwalters:
Part of https://github.com/coreos/rpm-ostree/issues/4701

Basically we should support setting arbitrary container configuration.


jmarrero (approved):
/lgtm

PR #568, opened 2023-11-28
various dependency bumps + lib: Release 0.12.7

cgwalters:
lib: Update to libsystemd 0.7

Just keeping up with things.

---

lib: Update to async-compression 0.4

Just keeping up with things.

---

lib: Update indoc to 2

Just keeping up with things.

---

lib: Update io-lifetimes to 2

Just keeping up with things.

---

lib: Drop unused bitflags dependency

We weren't using this.

---

lib: Release 0.12.7

One API improvement, plus various dependency bumps.

---


jmarrero (approved):
lgtm

PR #569, opened 2023-11-30
Move /var content to /usr/share/factory/var

cgwalters:
Instead of dropping `/var` content on the floor, move it into `/usr/share/factory/var`.

With this change, a systemd-tmpfiles fragment like this: `C+ /var - - - - -`
is sufficient to get "populate /var" automatically on initial installation and across upgrades.

### DIFF
--- a/lib/src/tar/write.rs
+++ b/lib/src/tar/write.rs
@@ -129,11 +129,16 @@ fn normalize_validate_path(path: &Utf8Path) -> Result<NormalizedPathResult<'_>> 
         if !found_first {
             if let Utf8Component::Normal(part) = part {
                 found_first = true;
-                // Now, rewrite /etc -> /usr/etc, and discard everything not in /usr.
                 match part {
+                    // We expect all the OS content to live in usr in general
                     "usr" => ret.push(part),
+                    // ostree has special support for /etc
                     "etc" => {
                         ret.push("usr/etc");
+                    }
+                    // Content in /var will get copied by a systemd tmpfiles.d unit
+                    "var" => {
+                        ret.push("usr/share/factory/var");
                     }
                     o => return Ok(NormalizedPathResult::Filtered(o)),
                 }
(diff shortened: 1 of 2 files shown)


cgwalters:
Pairs with https://github.com/ostreedev/ostree/pull/3103

jmarrero (approved):

mrguitar:
Will this also solve the problem of rpms installing under /opt? ...he asked while making doe eyes at @cgwalters  
On the surface it seems like it would.

cgwalters:
> Will this also solve the problem of rpms installing under /opt?

The core issue remains that:

- rollbacks won't roll back data in `/var`
- Content that is added is never automatically deleted; so you can "leak" state

But yes, some will probably start working modulo those caveats.

travier:
From https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html (see https://github.com/ostreedev/ostree/pull/3103):

```
C, C+
           Recursively copy a file or directory, if the destination files or directories do not exist yet
           or the destination directory is empty. Note that this command will not descend into
           subdirectories if the destination directory already exists and is not empty, unless the action
           is suffixed with +. Instead, the entire copy operation is skipped. If the argument is omitted,
           files from the source directory /usr/share/factory/ with the same name are copied. Does not
           follow symlinks. Contents of the directories are subject to time-based cleanup if the age
           argument is specified.
```

If I understand correctly, I've found the following caveats to document:

On updates (N -> N+1), the following will happen:

- Files added in N+1 are copied to /var.
- Files removed in N+1 are not removed from /var. New installations will not have those files.
- Files modified in N+1 are not copied to /var. New installations will have the N+1 version of those files.

On rollbacks (N+1 -> N):
- Files added in N+1 are left in /var.
- Files removed in N+1 are re-added to /var if something else removed them from /var in N+1.

In general, I think we should recommend users include another unit that calls to rsync (for example) with a list of files to update in sync between updates on boot. If they want files to be removed, they will have to make another unit.

I still need to verify that `C+` does not overwrite existing files on update. This is unclear from the man page.

PR #570, opened 2023-12-08
Support non-`/usr` content if ostree configured with overlayfs

cgwalters:
This pairs with https://github.com/ostreedev/ostree/pull/3114

Basically if ostree has an overlayfs (whether writable or a readonly composefs), we're safe to handle non-`/usr` content.

However, we still *always* drop content from the
"API filesystems" like `/run`, `/dev` etc.  Because really, no one should be adding stuff there.


cgwalters:
This obviates https://github.com/ostreedev/ostree/pull/2681 for a container-oriented flow.

jmarrero (approved):
lgtm

PR #575, opened 2024-01-03
lib: Remove `is_layered`

cgwalters:
In the container flow we *always* create a merge commit nowadays, and we generally always expect multiple layers.

The `is_layered` here used to be meaningful when it counted non-ostree layers, but nowadays it's always true.


jmarrero (approved):
lgtm

PR #577, opened 2024-01-03
lib: Fix omitted lifetime in associated consts

cgwalters:
See the tracking issue https://github.com/rust-lang/rust/issues/115010

There's now a warning for this in newer Rust versions.


jmarrero (approved):
lgtm

PR #578, opened 2024-01-03
store: Remove deprecated `query_image`, rename replacement

cgwalters:
Knocking one thing off the semver bump TODO.


jmarrero (approved):
lgtm

PR #579, opened 2024-01-03
store: Cleanup `copy` function

cgwalters:
store: Remove deprecated `copy` function

Part of semver break cleanups.

---

store: Rename `copy_as` -> `copy`

Now that the old one is removed.

---


jmarrero (approved):
lgtm

PR #580, opened 2024-01-03
Remove ExportLayout

cgwalters:
We don't support the v0 layout for a long time, this is the last remnant.

Part of semver breaking cleanups.


jmarrero (approved):
lgtm

Issue #582, opened 2024-01-05
http_proxy et al no longer passed to skopeo


srd424:
#566 added `--reset-env` to the setpriv call that invokes skopeo, which means information about proxies is no longer passed through.

As a totally horrible hack, putting this in `/usr/local/bin/skopeo` works:
```
#! /bin/bash

if [ $(systemctl whoami) = "rpm-ostreed.service" ]; then
	export http_proxy=http://nuc-cache-ostree.lan:3128
	export https_proxy=http://nuc-cache-ostree.lan:3128
fi

/usr/bin/skopeo "$@"
```
`setpriv` doesn't seem to support to whitelist env vars - perhaps they could be extracted from the environment before calling it, then passed to `env`?

PR #583, opened 2024-01-17
isolation: Only drop a few strategic environment variables

cgwalters:
Closes: https://github.com/ostreedev/ostree-rs-ext/issues/582


jmarrero (approved):
lgtm

PR #584, opened 2024-01-17
commit: Drop erroring on `/var` content

cgwalters:
Since https://github.com/ostreedev/ostree-rs-ext/pull/569/commits/b5cdaf44b30814eaa5c54aa44639d52b8034c72e we are going to try really hard to just Deal With whatever content we find in `/var`.  The semantics are somewhat suboptimal, but allow people to do somewhat suboptimal things instead of not being able to do container-image based updates at all.


jmarrero (approved):
:pray:

PR #585, opened 2024-01-18
ocidir: Change `create` -> `ensure`

cgwalters:
It's handy to have this be idempotent; we previously said "must be empty" but that wasn't true.


jmarrero (approved):
lgtm

PR #588, opened 2024-01-24
Release 0.13.1

jmarrero:
Mainly to pickup #583

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.13.0"
+version = "0.13.1"
 rust-version = "1.70.0"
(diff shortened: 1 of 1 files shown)


lib/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0"
 name = "ostree-ext"
 readme = "README.md"
 repository = "https://github.com/ostreedev/ostree-rs-ext"
-version = "0.13.0"
+version = "0.13.1"
```
cgwalters, line 9: So this is actually the first release of the 0.13 (semver incompat) series, so we don't need to bump the version.

I wanted to get in https://github.com/ostreedev/ostree-rs-ext/pull/589 before we officially release 0.13.0 too.

cgwalters:
I did a 0.13.0 release...however, we may also think about creating a release-0.12 branch to cherry pick fixes to that e.g. rpm-ostree could consume without porting too?    Maybe it's not worth it.

PR #586, opened 2024-01-18
container: Switch to cap-std when exporting to oci

cgwalters:
Prep for further work.

PR #593, opened 2024-01-25
encapsulate: Fix regression with relative directories

cgwalters:
https://github.com/ostreedev/ostree-rs-ext/pull/586/commits/b104a467e9ed06663194c3b60334a5246eb988fb broke the rpm-ostree test suite, which encapsulates to a *relative* file path and so us doing `chdir` broke it.

Switch to passing down the dirfd and using `/proc/self/fd`.

(Mutable global state in `chdir` is bad)


jmarrero (approved):
lgtm

PR #597, opened 2024-02-07
ci: Add reverse dependency testing targeting bootc

cgwalters:
Similar to https://github.com/ostreedev/ostree/pull/3152/commits/3623f0d80564f228d009df22fe32f97c848691c1 this patches bootc with this code, and runs a bootc install to get a bit more end-to-end testing.


jeckersb (approved):
lgtm

PR #599, opened 2024-02-09
tar: Generalize a TarImportConfig

cgwalters:
I plan to add another variant of this behavior in the future, and having a proper structure is better than a single `bool`.


jmarrero (approved):
lgtm

cgwalters:
Merged in https://github.com/ostreedev/ostree-rs-ext/pull/602

PR #600, opened 2024-02-09
tests: Strengthen check for /usr/share/factory/var logic

cgwalters:
Let's verify the files ended up in the expected place, not just that we had no content import warnings.


jmarrero (approved):
lgtm

PR #601, opened 2024-02-09
Bump MSRV to 1.74

cgwalters:
We really really want the MSRV-aware resolver, because we're just getting dragged on this by clap-lex bumping, when we don't have any dependency on newer versions of that really.


jmarrero (approved):
lgtm

PR #602, opened 2024-02-09
store: If ostree >= 2024.3, retain content in `/var`

cgwalters:
This is intended to pair with https://github.com/ostreedev/ostree/pull/3166

If we detect a new enough ostree version, then by default don't remap
content in `/var`, assuming that ostree itself will handle it.

In order to unit test this (without depending on the ostree version
that happens to be on the host) add an API to the importer
which allows overriding the version.

---

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -166,6 +166,8 @@ pub struct ImageImporter {
     disable_gc: bool, // If true, don't prune unused image layers
     /// If true, require the image has the bootable flag
     require_bootable: bool,
+    /// If true, we have ostree v2024.3 or newer.
+    ostree_v2024_3: bool,
     pub(crate) proxy_img: OpenedImage,
 
     layer_progress: Option<Sender<ImportProgress>>,
@@ -471,6 +473,7 @@ impl ImageImporter {
             proxy_img,
             target_imgref: None,
             no_imgref: false,
+            ostree_v2024_3: ostree::check_version(2024, 3),
             disable_gc: false,
             require_bootable: false,
             imgref: imgref.clone(),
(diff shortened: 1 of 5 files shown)


lib/src/container/store.rs
```diff
@@ -471,6 +473,7 @@ impl ImageImporter {
             proxy_img,
             target_imgref: None,
             no_imgref: false,
+            ostree_v2024_3: ostree::check_version(2024, 3),
```
cgwalters, line 476: Now of course, we in theory need to care about the case where a *new* ostree is deploying an old target.  But personally, I don't care too much about this; we'll get the new ostree out pretty quickly.  And also the emphasis on the bootc side is on self-installing operating systems, which don't have this problem.

jmarrero: mmm does this mean we need to make sure not to ship 2024.3 in any version of OCP that has 2024.2 or older and commit only to back ports?

jmarrero (approved):
lgtm

lib/src/container/store.rs
cgwalters: This is a good topic to raise.  There's a lot of complexity and nuance here.  First, all of this code is really only relevant in the case of deploying a container image with content in `/var`.

Our stock FCOS/RHCOS images don't have any.  Now until the *previous* change here in https://github.com/ostreedev/ostree-rs-ext/pull/569 we just dropped all container `/var` content on the floor.

This change is taking it from using systemd-tmpfiles with `C+` where we have "add new files" semantics to "at most once" semantics.

Now that all said:

> Now of course, we in theory need to care about the case where a new ostree is deploying an old target. 

For OCP (and really most cases) we are always going from old :arrow_right: new, as one would expect.  

So I don't think this is a problem.

The bigger thing to do here is to verify that these new semantics for `/var` work as expected before we do another release.

PR #594, opened 2024-01-31
deploy: Don't enforce container sigpolicy by default, add `--enforce-container-sigpolicy`

cgwalters:
This mirrors the change done in
https://github.com/containers/bootc/pull/230/commits/20d2d6cdc1bb2ddf7f8a6884ad83efaa708b4c65

Basically our attempts to enforce this didn't really work out.

PR #604, opened 2024-02-13
container: Allow alternative formatting to drop `ostree-unverified-registry`

cgwalters:
Since
https://github.com/ostreedev/ostree-rs-ext/pull/594/commits/34fea14d5b6dd4b24919d4643b2c86c01273d25c we are backing off the signature verification bits.  Let's make it easier to not put those bits in the user's face by adding an alternative formatting that drops it if it's the default.


jmarrero (approved):
lgtm

PR #605, opened 2024-02-29
lib/tar: Add much more error context

cgwalters:
It turns out that the error we were seeing is that inside osbuild it wasn't mounting `/var/tmp`, and the tmpdir allocation was failing.  This was not an obvious error, so add a bunch of error contexts.


jeckersb (approved):
Big :+1: (probably don't have approval access though?)

jeckersb:
Yay I do afterall! :smile:

cgwalters:
Followup in https://github.com/ostreedev/ostree-rs-ext/pull/606

PR #606, opened 2024-02-29
tar: Unconditionally use repo tmpdir

cgwalters:
The tar code had some fancy logic to lazily allocate a temporary directory if it turned out we needed one, which only broke in the rare case we needed one in an obscure circumstance (a bwrap container in osbuild).

But we already always have a tmpdir allocated in the ostree repo, so switch the code to use that.


jmarrero (approved):
lgtm

Issue #608, opened 2024-03-05
Add support for zstd:chunked


cgwalters:
Should be a trivial patch

travier:
From "rpm-ostree":
```
error: Importing: Unencapsulating base: Unhandled layer type: application/vnd.oci.image.layer.v1.tar+zstd
```

travier:
Unfortunately https://github.com/ostreedev/ostree-rs-ext/pull/615 is for zstd only, not zstd:chunked which is not supported by the [oci_spec library yet](https://docs.rs/oci-spec/latest/oci_spec/image/enum.MediaType.html#variants).

travier:
For reference: https://github.com/containers/storage/pull/775

jmarrero:
> From "rpm-ostree":
> 
> ```
> error: Importing: Unencapsulating base: Unhandled layer type: application/vnd.oci.image.layer.v1.tar+zstd
> ```

I am seeing this on OCP 4.16 is there a fix that needs to be backported?

jmarrero:
Also are we building images with  application/vnd.oci.image.layer.v1.tar+zstd support? without supporting it yet on OCP?

travier:
Looks like podman started pushing using zstd compression by default (?). We only have support for that in the latest rpm-ostree release AFAIK.

cgwalters:
> I am seeing this on OCP 4.16 is there a fix that needs to be backported?

Where?  In Prow CI?  Local builds?  A customer bug?

> Looks like podman started pushing using zstd compression by default (?).

Wait, really?  That's going to cause some fallout for sure.  I think this is probably from https://bodhi.fedoraproject.org/updates/FEDORA-2024-ab42dd0ffb

jmarrero:
> > I am seeing this on OCP 4.16 is there a fix that needs to be backported?
> 
> Where? In Prow CI? Local builds? A customer bug?
> 
Using the clusterbot created cluster OCP 4.16-rc3 I was trying to help verify an issue that includes replacing the kernel and hit this with coreos layering. 

I did build the image I am rebasing to using Fedora 40.

cgwalters:
OK, in the meantime, the workaround for this is probably to do `skopeo copy --dest-compress-format=gzip containers-storage:<local image> docker://quay.io/exampleuser/someuser`

Or set `compression_format = "gzip"` in `/etc/containers/containers.conf` (locally) or revert the change in the `/usr` copy in custom images.

[...]

PR #615, opened 2024-03-27
Add zstd:chunked support

travier:
Fixes: https://github.com/ostreedev/ostree-rs-ext/issues/608


travier:
Builds but completely untested thus the draft.

cgwalters (approved):
Thanks for doing this!  I added a test case, let's merge!

PR #618, opened 2024-04-16
container: Support JSON out for ManifestDiff

travier:
This will let higher level tools (Plasma Discover for example) more easily read the output of `bootc update --check --json`.


travier:
Draft as we probably want to use a JSON crate for this.

travier:
c9s test is failing because it tries to build bootc and thus needs https://github.com/containers/bootc/pull/472.

Maybe I should not change the number of args of this function and add a new `print_json` one instead?

lib/src/container/mod.rs
```diff
@@ -376,16 +377,34 @@ impl<'a> ManifestDiff<'a> {
...
-        let print_n_removed = self.n_removed;
-        let print_removed_size = glib::format_size(self.removed_size);
-        let print_n_added = self.n_added;
-        let print_added_size = glib::format_size(self.added_size);
-        println!("Total new layers: {print_total:<4}  Size: {print_total_size}");
-        println!("Removed layers:   {print_n_removed:<4}  Size: {print_removed_size}");
-        println!("Added layers:     {print_n_added:<4}  Size: {print_added_size}");
+    pub fn print(&self, json: bool) {
+        if json {
+            let out = json!({
```
cgwalters, line 382: Instead of this we can add `#[derive(Serialize)]` for ManifestDiff and get these bits for free.  We'll likely need to skip serializing the held references.

cgwalters:
Closing in favor of https://github.com/ostreedev/ostree-rs-ext/pull/619

lib/src/cli.rs
```diff
@@ -234,6 +239,10 @@ pub(crate) enum ContainerImageOpts {
         /// the new manifest.
         #[clap(long)]
         check: Option<Utf8PathBuf>,
```
travier, line 241: It looks like this Option is not respected by clap and always asked on the command line:

```
$ ostree container image pull --check check repo ostree-image-signed:registry:a.a/a:a
error: Opening directory at 'repo': No such file or directory (os error 2)

$ ostree container image pull --check repo ostree-image-signed:registry:a.a/a:a
error: the following required arguments were not provided:
  <IMGREF>

Usage: ostree-container container image pull --check <CHECK> <REPO> <IMGREF>

For more information, try '--help'.
```

travier: Hum, OK, it's a little bit more complex than that for the pull --check option as it optionally writes to a file.

PR #619, opened 2024-04-17
container: Make ManifestDiff serializable

cgwalters:
But don't serialize the layers or manifests as that's large; just the numeric diff which is a lot smaller amount of data.

Fetching manifests and layers should be done via a dedicated API.


travier (approved):
LGTM. I don't think that cover the full scope of https://github.com/ostreedev/ostree-rs-ext/pull/618 however?

cgwalters:
Yeah you're right, reopened that.

PR #620, opened 2024-04-26
Add docker-archive transport

albertofaria:
This would help enable [crun-vm](https://github.com/containers/crun-vm)'s support for [bootc](https://github.com/containers/bootc/) containers when running under Docker.


cgwalters (approved):
Thanks!

PR #621, opened 2024-04-28
wip: attempt to drop async_compression

cgwalters:
This would be a workaround for the async_compression bug in https://github.com/ostreedev/ostree-rs-ext/issues/608

- This changes the public API so we need a semver bump
- It's pretty risky anyways and we may be better off with spawning a helper thread instead of using async_compression


cgwalters:
https://github.com/ostreedev/ostree-rs-ext/pull/622

PR #625, opened 2024-05-21
Fix use of deprecated chrono function

jeamland:
This creates a `DateTime<Utc>` rather than a `NaiveDateTime` but it's what chrono recommends and, based on examining the usage of the object, doesn't change any outcomes.


jeckersb (approved):
Thanks for this!

I have an identical change sitting in an unpublished branch along with some other clippy lint fixes, this will motivate me to rebase my changes on top of this and finally submit them :smile:

PR #628, opened 2024-05-21
lib: use pkg-config feature for zstd

cgwalters:
I don't want to default to using vendored C sources.


jeckersb (approved):
:+1: good catch!

jeckersb:
Hm need to update deps to include libzstd-devel for this.  Added:

- https://src.fedoraproject.org/rpms/bootc/pull-request/3
- https://github.com/containers/bootc/pull/560

jeckersb:
Also https://github.com/ostreedev/ostree-rs-ext/pull/629

PR #629, opened 2024-05-22
installdeps: Add libzstd-devel

jeckersb:


jeckersb:
Actually it looks like this is already included in the fcos container, so not needed.

PR #631, opened 2024-05-23
lib: Re-export containers_image_proxy in full

cgwalters:
We currently expose a type from that, so we might as well just re-export the whole thing.

I want to do some stuff directly with containers in bootc.


HuijingHei (approved):
LGTM

Issue #633, opened 2024-05-29
ostree container image deploy don't support `composefs=signed`


jbtrystram:
While working on a POC enabling a signed composeFS in coreos-assembler I get the following error at the `ostree container image deploy` step: 

```
error: Reading composefs config: Loading composefs config: Invalid tri-state value: signed
```

See https://github.com/coreos/coreos-assembler/pull/3813#issuecomment-2137773459 for more details

EDIT: Ah, sorry I should have opened the issue against the ostree repo

cgwalters:
> EDIT: Ah, sorry I should have opened the issue against the ostree repo

This is the correct repo actually because this code has a different parser it's broken here

PR #635, opened 2024-05-29
prepareroot: Also accept composefs enabled=signed

cgwalters:
This was an oversight, unfortunately having two different programming languages here led to less code sharing.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/633

### DIFF
--- a/lib/src/ostree_prepareroot.rs
+++ b/lib/src/ostree_prepareroot.rs
@@ -84,6 +84,39 @@ impl Tristate {
     }
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/ostree_prepareroot.rs
```diff
@@ -112,6 +145,27 @@ fn test_tristate() {
     }
 }
 
+#[test]
+fn test_composefs_state() {
```
jeckersb, line 149: :heart:

jeckersb (approved):
:+1:

PR #636, opened 2024-05-31
Make authfile API public

cgwalters:
Prep for having bootc directly read these as part of using podman for pulls.

While we're here:

- Use cap-std on general principle
- Change the usage to just honoring the file content directly instead of re-reading the file
- Add unit tests using that too
- Switch to using camino for paths

### DIFF
--- a/lib/src/container/mod.rs
+++ b/lib/src/container/mod.rs
@@ -26,6 +26,8 @@
 //! for this is [planned but not implemented](https://github.com/ostreedev/ostree-rs-ext/issues/12).
 
 use anyhow::anyhow;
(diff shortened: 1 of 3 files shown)


jeckersb (approved):
:+1: 

I added an extra commit with tests for the non-root case

cgwalters:
> I added an extra commit with tests for the non-root case

Awesome, thanks!  I hadn't thought about unit testing those by stripping the absolute path in the tests, makes sense.

cgwalters:
Regression in https://github.com/containers/bootc/issues/679

PR #638, opened 2024-06-03
Update to use external `ocidir` crate

cgwalters:
I moved the code that was in `ocidir` to a new repository and crate: https://github.com/containers/ocidir-rs/

It definitely cleans things up here.


cgwalters:
Ah right, this is also a semver break because we're bumping `oci-spec`, which is why the bootc tests fail.

https://github.com/containers/containers-image-proxy-rs/pull/65

jeckersb (approved):

PR #639, opened 2024-06-05
proxy: Fix logic error with auth_data switch

cgwalters:
The previous change switched to use `auth_data` directly instead of passing the file.  Update the logic just below which needs adjustment to continue to support authenticated pulls.


cgwalters:
Will look at some test coverage for this using ghcr.io per PR

jeckersb (approved):

PR #641, opened 2024-06-19
Container export prep

cgwalters:


cgwalters:
Eh I rolled this into https://github.com/ostreedev/ostree-rs-ext/pull/642

PR #642, opened 2024-06-19
container: Add support for re-exporting a fetched container

cgwalters:
chunking: Add some doc comments

Just a drive by.

---

container: Add support for re-exporting a fetched container

The status quo today is basically that with a "pure ostree"
container image, one can pull it, and one *can* re-export
it with `ostree container encapsulate`...but doing so
loses *all chunking* i.e. you end up with a single
giant layer again.

Further, we don't support exporting derived images at all.

Fix both of these with a new CLI and API, for example:

```
$ ostree container image export --repo=/path/to/repo registry:quay.io/exampleos/someos:latest containers-storage:localhost/exampleos
```

Now...before one gets too excited, this is still suboptimal
in a bunch of ways:

- Copying to `containers-storage` is super inefficient, we indirect
  through a local `oci` directory because of the lack of "push"
  support in containers-image-proxy, *and* we end up with a full
  physical copy of the files even when we *could* reflink;
  cc https://github.com/containers/storage/issues/1849
- Because we don't currently save tar-split data, the use
  case of pushing to a registry is virtually guaranteed to produce
  changed diffids, and we'll hence end up duplicating layers
  on the registry

Now what is more interesting is that this code is going to help
us a bit for the use case of "recommitting" a derived container
image.

Signed-off-by: Colin Walters <walters@verbum.org>

---

### DIFF
--- a/lib/tests/it/main.rs
+++ b/lib/tests/it/main.rs
@@ -10,9 +10,9 @@ use ostree_ext::container::{store, ManifestDiff};
 use ostree_ext::container::{
     Config, ExportOpts, ImageReference, OstreeImageReference, SignatureSource, Transport,
 };
-use ostree_ext::ostree_manual;
 use ostree_ext::prelude::{Cast, FileExt};
 use ostree_ext::tar::TarImportOptions;
+use ostree_ext::{fixture, ostree_manual};
 use ostree_ext::{gio, glib};
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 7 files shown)


lib/tests/it/main.rs
```diff
@@ -601,6 +601,99 @@ async fn impl_test_container_import_export(chunked: bool) -> Result<()> {
...
+async fn test_export_as_container_derived() -> Result<()> {
+    let fixture = Fixture::new_v1()?;
+    // Export into an OCI directory
+    let src_imgref = fixture.export_container().await.unwrap().0;
+    // Add a derived layer
+    let derived_tag = "derived";
+    // Build a derived image
+    let srcpath = src_imgref.name.as_str();
+    let temproot = &fixture.path.join("temproot");
+    || -> Result<_> {
```
jmarrero, line 648: This looks like magic to me... the examples I see of closures assign to a variable. Is this using a closure to use the variables you just declared before instead of writing a function and passing variables to it?

jmarrero (approved):
lgtm

lib/tests/it/main.rs
cgwalters: BTW, totally fine asking questions like this on PRs, please continue to do so!

This is an [IIFE](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) - it's...somewhat unusual. The rationale here is actually so that `    .context("generating temp content")?;` applies to *everything* inside the closure - it's a way of adding context to a whole bunch of `?` without actually splitting it out into a separate function as you say.

Splitting into a separate function is more heavyweight: requires spelling out all the types and passing all the inputs.

But yes, I don't see IIFEs in much other Rust code, and I don't tend to use them often.  If I remember right, I added this one because I had a bug *somewhere* in all those lines and I didn't want to change everything to use `unwrap()`.

jmarrero: thanks that makes sense.

cgwalters: That said, I actually copy-pasted this, so https://github.com/ostreedev/ostree-rs-ext/pull/643 dedups it

PR #644, opened 2024-06-21
export: Add progress flag

cgwalters:
This is a very slow operation and it'd be really
nice to have progress.

Unfortunately we don't have a sane way to propagate status from `skopeo copy`, so add an option to just output to stdout, which is usable now in bootc.


jeckersb (approved):
LGTM

PR #648, opened 2024-06-30
tar: Handle hardlinks into /etc

cgwalters:
Because we rewrite `etc` -> `usr/etc`, we must also rewrite hardlinks.


cgwalters:
Workaround example https://github.com/containers/ai-lab-recipes/pull/650/commits/0b5b374537b90695125f2361559a13021d20634e

jeckersb (approved):

PR #650, opened 2024-07-15
config: Fix absolute symlinks

cgwalters:
Use the API from https://github.com/coreos/cap-std-ext/pull/54 to fix absolute symlinks for `/etc/ostree/auth.json`.

cc https://github.com/containers/bootc/issues/679


cgwalters:
Now using released cap-std-ext

jmarrero (approved):
lgtm

PR #656, opened 2024-08-29
Require matching architecture for bootable images

cgwalters:
xref https://github.com/containers/bootc/issues/682

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -20,7 +20,9 @@ use containers_image_proxy::{ImageProxy, OpenedImage};
 use flate2::Compression;
 use fn_error_context::context;
 use futures_util::TryFutureExt;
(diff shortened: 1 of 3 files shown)


mvo5:
Just to double check my understanding: `bootc-image-builder --target-arch` is not affected here because bib does the foreign arch container pulling and when we run bootc it is already a "foreign" arch binary(?)

cgwalters:
> Just to double check my understanding: bootc-image-builder --target-arch is not affected here because bib does the foreign arch container pulling and when we run bootc it is already a "foreign" arch binary(?)

That sounds right yes. (Using the term both "target" and "foreign" here is a bit confusing; I think we can just say: bib pulls the image with the target arch, and when we run bootc it's running emulated with the target arch, so this check will pass)

jmarrero (approved):
lgtm

PR #658, opened 2024-08-30
lib: fix building with musl libc

JustSoup312:
No major changes, just defines a const that is unavailable from the libc crate under musl.

libc does not provide STATX_ATTR_MOUNT_ROOT for the musl target as the integration for it has been slow at best. Various dependent projects have included this const manually until the libc crate itself fixes the issues with musl support.

This shouldn't change compilation on glibc-based distros as conditional-compilation is invoked.

### DIFF
--- a/lib/src/mountutil.rs
+++ b/lib/src/mountutil.rs
@@ -7,12 +7,18 @@ use anyhow::Result;
 use cap_std::fs::Dir;
 use cap_std_ext::cap_std;
 
+// Fix musl support
+#[cfg(target_env = "gnu")]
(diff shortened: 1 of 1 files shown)


lib/src/mountutil.rs
```diff
@@ -7,12 +7,18 @@ use anyhow::Result;
 use cap_std::fs::Dir;
 use cap_std_ext::cap_std;
 
+// Fix musl support
```
cgwalters, line 10: I'm fine with this as is but probably the best place to fix this is adding the definitions into rustix

cgwalters (approved):

PR #663, opened 2024-09-18
Port to ocidir 0.3, containers-image-proxy 0.7

jeckersb:

### DIFF
--- a/lib/src/container/unencapsulate.rs
+++ b/lib/src/container/unencapsulate.rs
@@ -101,18 +101,18 @@ impl<T: AsyncRead> AsyncRead for ProgressReader<T> {
 async fn fetch_manifest_impl(
     proxy: &mut ImageProxy,
     imgref: &OstreeImageReference,
-) -> Result<(oci_spec::image::ImageManifest, String)> {
+) -> Result<(oci_image::ImageManifest, oci_image::Digest)> {
     let oi = &proxy.open_image(&imgref.imgref.to_string()).await?;
     let (digest, manifest) = proxy.fetch_manifest(oi).await?;
     proxy.close_image(oi).await?;
-    Ok((manifest, digest))
+    Ok((manifest, oci_image::Digest::from_str(digest.as_str())?))
 }
 
 /// Download the manifest for a target image and its sha256 digest.
 #[context("Fetching manifest")]
 pub async fn fetch_manifest(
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -17,3 +17,7 @@ debug = true
 codegen-units = 1
 inherits = "release"
 lto = "yes"
+
+[patch.crates-io]
```
jeckersb, line 21: Note the explicit patch to align with the containers-image-proxy change here, need to coordinate moving these in tandem.

cgwalters:
Ah sorry, my porting to `thiserror` broke this one. I think we may just need a few `.map_err(Into::into)`?

jeckersb:
> Ah sorry, my porting to `thiserror` broke this one. I think we may just need a few `.map_err(Into::into)`?

Holy cow is that an absolutely cryptic compiler error now :laughing:

lib/src/container/unencapsulate.rs
```diff
@@ -289,19 +290,17 @@ pub(crate) async fn fetch_layer_decompress<'a>(
...
             size = layer.size();
             media_type = layer.media_type();
-            (blob, driver) = proxy
-                .get_blob(img, layer.digest().as_str(), size as u64)
-                .await?;
+            (blob, driver) = proxy.get_blob(img, layer.digest(), size).await?;
         }
     };
 
+    let driver = async { driver.await.map_err(Into::into) };
```
jeckersb, line 302: seems to be enough to accomodate the `thiserror` change, took me longer than I would care to admit to understand the compiler error before.

cgwalters: I struggled a lot with this when I was learning Rust for sure. The thing that is subtle is that the `?` operator automatically also does `Into::into`, which means in cases where you're changing the error type (like this) but *not* using `?` (which happens in most cases) you suddenly have to think about these conversions.

cgwalters:
Ah, test run hit https://github.com/ostreedev/ostree-rs-ext/issues/664

cgwalters (approved):
LGTM but we also need to remember this is a semver-breaking change for us, so we'll need to bump the version for the next release.

cgwalters:
Fallout in https://github.com/containers/bootc/issues/800

I have yanked version 0.15.0.

PR #666, opened 2024-09-23
Fix digest usage

cgwalters:
In some places we started doing `.digest().digest()` which drops the algorithm. Some basic unit tests
would have caught this - I added one. We also need an "upgrade from previous" test.

Trying to fix this revealed a bunch more places where we had `String` and not `Digest`.

Fixing those gets us type safety and points out the problem areas.  Since we're breaking API as this will be the real first 0.15 release,

- Drop unnecessary ManifestLayerState wrappering
- Fix the manifest digest members of various structs to be a Digest

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -21,7 +21,7 @@ use flate2::Compression;
 use fn_error_context::context;
 use futures_util::TryFutureExt;
 use oci_spec::image::{
-    self as oci_image, Arch, Descriptor, History, ImageConfiguration, ImageManifest,
+    self as oci_image, Arch, Descriptor, Digest, History, ImageConfiguration, ImageManifest,
 };
 use ostree::prelude::{Cast, FileEnumeratorExt, FileExt, ToVariant};
 use ostree::{gio, glib};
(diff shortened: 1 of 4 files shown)


cgwalters:
> .digest().digest()

In other words this one should be a big red flag in code - it's never right to just ignore the algorithm. What's the most common though is a pattern like:

- Assert that the algorithm is sha256
- If so return that digest and use it in a place that wants "plain" sha256 without the algorithm prefix

We could look at adding a helper for that.

lib/src/container/store.rs
```diff
@@ -1808,3 +1805,26 @@ pub(crate) fn verify_container_image(
...
 }
+
+#[cfg(test)]
+mod tests {
+    use oci_image::{DescriptorBuilder, MediaType, Sha256Digest};
+
+    use super::*;
+
+    #[test]
+    fn test_ref_for_descriptor() {
```
jeckersb, line 1816: :heart:

lib/src/container/store.rs
```diff
@@ -21,7 +21,7 @@ use flate2::Compression;
 use fn_error_context::context;
 use futures_util::TryFutureExt;
 use oci_spec::image::{
-    self as oci_image, Arch, Descriptor, History, ImageConfiguration, ImageManifest,
+    self as oci_image, Arch, Descriptor, Digest, History, ImageConfiguration, ImageManifest,
```
jeckersb, line 24: We still use `oci_image::Digest` in a bunch of places, not a big deal but I can do a followup PR after this merges just so we're consistent everywhere.

jeckersb (approved):
LGTM.  I definitely relied entirely too much on the compiler and tests to sanity check the original patch.  Some of this is really obvious in retrospect!

jeckersb:
Obligatory acknowledgement that the bootc build test is failing because we need to do the CI dance to get both updated in-tandem.

Issue #510, opened 2023-07-30
OCI SELinux labeling mismatch when package only ships binary policy - greetd is broken


akdev1l:
Hi OSTree team,

I am trying to debug an issue that surfaced in Ublue (https://github.com/ublue-os/main/issues/223). In Ublue we are making use of the OCI functionality heavily to produce custom downstream images of Fedora Silverblue. Some users have wanted to install `greetd` as their greeter for simplicity but whenever they do the resulting image is slightly broken because of wrong SELinux labeling.

I have investigated the package and I see that `/usr/bin/greetd` should be labeled as `xdm_exec_t`(see: [here](https://src.fedoraproject.org/rpms/greetd/blob/rawhide/f/greetd.fc)) which makes sense for a display manager but in the resulting images we see that `/usr/bin/greetd` is labeled as `bin_t` and therefore access gets denied by SELinux when the system is booted.

This is the output of `ls -lZ` in the images:

```
-rwxr-xr-x. 3 root root system_u:object_r:bin_t:s0 839488 Jan  1  1970 /usr/bin/greetd
```

this is the package source that shows the intended labeling: 

I tried to trace down the issue and I found this function: `sepolicy_from_base`:

```
  1 fn sepolicy_from_base(repo: &ostree::Repo, base: &str) -> Result<tempfile::TempDir> {
  2     let cancellable = gio::Cancellable::NONE;
  3     let policypath = "usr/etc/selinux";
  4     let tempdir = tempfile::tempdir()?;
  5     let (root, _) = repo.read_commit(base, cancellable)?;
  6     let policyroot = root.resolve_relative_path(policypath);
  7     if policyroot.query_exists(cancellable) {
  8         let policydest = tempdir.path().join(policypath);
  9         std::fs::create_dir_all(policydest.parent().unwrap())?;
 10         let opts = ostree::RepoCheckoutAtOptions {
 11             mode: ostree::RepoCheckoutMode::User,
 12             subpath: Some(Path::new(policypath).to_owned()),
 13             ..Default::default()
 14         };
 15         repo.checkout_at(Some(&opts), ostree::AT_FDCWD, policydest, base, cancellable)?;
 16     }
 17     Ok(tempdir)
 18 }
 ```
 
[...]

PR #669, opened 2024-09-30
store: Use policy from merged tree with derived layers

cgwalters:
I was initially worried this was going to slow things down but actually AFAICS because we are already setting a policy we end up rechecking the labels for all of the base image files anyways, even in the optimal
path.

So what's really happening here is we could likely speed up the non-derived case. But let's leave that as a TODO as we have much more important things.

Closes: https://github.com/ostreedev/ostree-rs-ext/issues/510

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -146,6 +146,24 @@ jobs:
         run: install ostree-ext-cli /usr/bin && rm -v ostree-ext-cli
       - name: Integration tests
         run: ./ci/priv-integration.sh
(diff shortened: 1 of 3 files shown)


cgwalters:
This works for me, but draft since it could use a test.

cgwalters:
Put up https://gitlab.com/fedora/bootc/tests/container-fixtures/-/merge_requests/11

cgwalters:
Now with a test case! (In shell script because...ug)

jmarrero (approved):
lgtm

PR #671, opened 2024-10-21
unencapsulate: Add errctx for decompression

cgwalters:
We think this may be implicated in the mysterious "broken pipe" errors.


jmarrero (approved):
lgtm

PR #673, opened 2024-10-24
Import generic prep

cgwalters:
tar: Don't filter out toplevel /run, /proc etc

Otherwise we break things when using this code to process
a "non-ostree" image.

Signed-off-by: Colin Walters <walters@verbum.org>

---

store: Accept just containers.bootc for as bootable

This allows us to accept images to that don't have any metadata
keys starting with `ostree`.

---

### DIFF
--- a/lib/src/container/encapsulate.rs
+++ b/lib/src/container/encapsulate.rs
@@ -27,6 +27,8 @@ pub const LEGACY_VERSION_LABEL: &str = "version";
 /// The label which indicates where the ostree layers stop, and the
 /// derived ones start.
 pub const DIFFID_LABEL: &str = "ostree.final-diffid";
+/// The label for bootc.
+pub const BOOTC_LABEL: &str = "containers.bootc";
 
 /// Annotation injected into the layer to say that this is an ostree commit.
 /// However, because this gets lost when converted to D2S2 https://docs.docker.com/registry/spec/manifest-v2-2/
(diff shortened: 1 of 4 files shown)


lib/src/container/encapsulate.rs
```diff
@@ -68,7 +70,7 @@ fn commit_meta_to_labels<'a>(
     #[allow(clippy::explicit_auto_deref)]
     if let Some(v) = meta.lookup::<bool>(*ostree::METADATA_KEY_BOOTABLE)? {
         labels.insert(ostree::METADATA_KEY_BOOTABLE.to_string(), v.to_string());
-        labels.insert("containers.bootc".into(), "1".into());
+        labels.insert(BOOTC_LABEL.into(), "1".into());
```
djach7, line 73: What is `v` in this context? Is it just a version number?

djach7 (commented):
Tested and built successfully for me, apart from that `test_tar_write` test failing. Are we ignoring that for now? Otherwise lgtm, I just had a clarifying question about `v`.

cgwalters:
Fixed the CI test

lib/src/container/encapsulate.rs
cgwalters: short for "value"

cgwalters: But it's always a boolean here

djach7: Ah, gotcha. Just couldn't figure out what it was short for.

cgwalters:
@jeckersb at the moment David doesn't have merge rights (i invited him) but can you stamp this?

djach7 (approved):
Just accepted the inv, lgtm

PR #674, opened 2024-10-24
store: Support importing images without `/ostree`

cgwalters:
A sticking point keeping ostree in the picture here for
containers was SELinux handling. When we started this
effort I'd feared rewriting.

But recently we changed things such that we label derived images
using the policy from the final root.

This is a relatively small change in code size and complexity,
that allows us to import images that don't have "ostree stuff"
in them at all, i.e. there's no `/ostree/repo/objects`.

The advantage here is that this significantly simplifies
constructing base images.

The main disadvantage today for people who build images
this way is that we end up re-labeling and re-checksumming all objects.

But, the real fix for that in the future will be for us to
rework things such that we support `security.selinux` for example
as native xattrs in the tar stream.


djach7 (approved):
Looks like the same testing failure as #673 but apart from that it builds and tests for me. LGTM but I'm going to keep reading and digging in because there's a lot here that I'm not familiar with.

cgwalters:
I'm picking this one back up, fixing conflicts and adding tests

cgwalters:
OK rebased :surfer: and now with tests!

However, something kind of left open here is basically nailing down things like:

- Do we require `ostree/` and `sysroot` to be in the image, or do we just...make them if they don't exist?

cgwalters:
OK for reference, this gets to running with `bootc install` and almost to booting for me with this patch (and https://github.com/containers/bootc/pull/844 ).

```
FROM quay.io/fedora/fedora:41
RUN <<EORUN

set -xeuo pipefail
dnf -y install rpm-ostree kernel systemd bootc bootupd ostree xfsprogs dosfstools \
               grub2 grub2-efi-x64 efibootmgr shim microcode_ctl \
               NetworkManager openssh nss-altfiles
dnf clean all

kver=$(cd /usr/lib/modules && echo *)
echo "kver=$kver"
# The kernel is duplicated in /boot, and the initramfs needs regenerating
rm -vf /boot/vmlinuz* /boot/initramfs*
dracut --no-hostonly /usr/lib/modules/$kver/initramfs.img "$kver"
mkdir -p /usr/lib/ostree-boot
mv /boot/* /usr/lib/ostree-boot
# FIXME drop this requirement for bootupd
ln -sr /usr/lib/sysimage/rpm /usr/share/rpm
bootupctl -vvvv backend generate-update-metadata
mkdir /sysroot
ln -s sysroot/ostree .
EORUN
LABEL containers.bootc=1
```

So we can use this as an optimization target.

cgwalters:
Although moving back to draft as this technically breaks semver, and I'd like to do another release with other fixes before we do this one.

Issue #664, opened 2024-09-19
test flake `error: opening repo: opendir(/proc/self/fd/3): No such file or directory`


cgwalters:
```
Caused by:
    0: Parsing layer blob 6d52d9af072841028a446e6a01f4a89484ae4ed93f235cfbccd22a877231736e
    1: error: opening repo: opendir(/proc/self/fd/3): No such file or directory
    2: Processing tar
    3: Failed to commit tar: ExitStatus(unix_wait_status(256))
```

We see this pretty often, it's a bit concerning as I have no idea how it could just be flaky.

cgwalters:
Put up https://github.com/ostreedev/ostree-rs-ext/pull/676

jeckersb:
I was able to get this to reproduce under `perf` in `test_tar_write_tar_layer`.  Since we always use fd=3 it was straightforward to just grab everywhere we call `close(3)` where the call stack is under `test_tar_write_tar_layer`.  Here's the places (duplicates omitted) where we end up dropping fd 3:

```
test_tar_write_  192137 [005]  6965.825170: probe:file_close_fd: (ffffffffb24d59c4) fd=0x3
        ffffffffb24d59c5 file_close_fd+0x5 ([kernel.kallsyms])
        ffffffffb24a4c93 __x64_sys_close+0x13 ([kernel.kallsyms])
        ffffffffb31b2952 do_syscall_64+0x82 ([kernel.kallsyms])
        ffffffffb320012f entry_SYSCALL_64+0xaf ([kernel.kallsyms])
            7f1838b17f9c __libc_close+0x3c (inlined)
            7f1839128bc3 _gpgme_io_spawn+0x533 (/usr/lib64/libgpgme.so.11.32.1)
            7f1839130434 _gpgme_get_program_version+0x104 (/usr/lib64/libgpgme.so.11.32.1)
            7f1839116cfa engine_get_version+0x9a (inlined)
            7f1839116cfa _gpgme_set_engine_info+0x9a (inlined)
            7f183912df69 _gpgme_set_engine_info+0x139 (inlined)
            7f183912df69 gpgme_ctx_set_engine_info+0x139 (/usr/lib64/libgpgme.so.11.32.1)
            7f1839730735 ot_gpgme_new_ctx+0x55 (/usr/lib64/libostree-1.so.1.0.0)
            7f183973cde0 sign_data.constprop.0.isra.0+0xc0 (/usr/lib64/libostree-1.so.1.0.0)
[...]

PR #676, opened 2024-10-31
tar/write: Ensure repofd outlives child spawn

cgwalters:
Might help with https://github.com/ostreedev/ostree-rs-ext/issues/664 But I didn't manage to reliably reproduce this.


jeckersb:
Doesn't appear to help at all :cry:

jeckersb:
FWIW I tried to move the drop all the way to the end until after we join with the child process, that also does not help.

PR #677, opened 2024-10-31
Remove decompress_bridge and move decompression inline

### LINKED ISSUE #657: Parsing layer blob: Broken pipe
> stderr: ERROR: Switching: Pulling: Importing: Parsing layer blob sha256:4367367aae6325ce7351edb720e7e6929a7f369205b38fa88d140b7e3d0a274f: Broken pipe (os error 32)"

This one is like my enemy!  I have a tracker over here for it https://github.com/coreos/rpm-ostree/issues/4567 too

---

Discoveries so far:

- https://github.com/ostreedev/ostree-rs-ext/pull/527 does not help (but doesn't hurt either)
- This test ends up pulling a 1.5G layer
- Retrying *does* work - so we're apparently writing the layer correctly

More generally it's definitely a race condition; I can sometimes reproduce this by doing
`ostree refs --delete ostree/container` and then re-running the rebase.

Also of note: kola defaults to a uniprocessor VM, which I think is more likely to expose this race.

I'm quite certain it has something to do with the scheduling of us closing the pipe vs calling `FinishPipe`.

jeckersb:
This (with cautious optimism) appears to fix #657.

### DIFF
--- a/lib/src/container/store.rs
+++ b/lib/src/container/store.rs
@@ -716,7 +716,7 @@ impl ImageImporter {
                 p.send(ImportProgress::OstreeChunkStarted(layer.layer.clone()))
                     .await?;
             }
-            let (blob, driver) = fetch_layer_decompress(
+            let (blob, driver, media_type) = fetch_layer(
                 &self.proxy,
                 &self.proxy_img,
                 &import.manifest,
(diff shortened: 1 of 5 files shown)


cgwalters (approved):

cgwalters:
The other CI thing going on here is we got dnf5 in fcos testing-devel which of course just broke super basic (IMO) CLI invocations...

cgwalters:
Let's get https://github.com/ostreedev/ostree-rs-ext/pull/678 in and then rebase this?

cgwalters:
Ok merged that, rebase?

jeckersb:
:crossed_fingers:

cgwalters:
Thanks for this!

PR #678, opened 2024-10-31
installdeps: drop dnf config-manager

cgwalters:
Because dnf5 broke it and keeping both working is annoying and we didn't really need it.


jeckersb (approved):

jeckersb:
Of course it hit #664 :roll_eyes:

cgwalters:
FTR I filed https://github.com/rpm-software-management/dnf-plugins-core/issues/565

PR #680, opened 2024-11-01
tar: Handle absolute hardlinked paths to sysroot

cgwalters:
See https://github.com/containers/bootc/issues/856

Basically we need to handle absolute paths here too.


jmarrero (approved):
lgtm

